Issue #2, opened 2018-03-22
Continuous Integration


mrocklin:
This repository could use continuous integration to ensure that the examples continue to work over time

I'm not familiar with tools to test notebooks, but other projects do this, so it must be doable.

mholtzscher:
@mrocklin do you have any example projects that do the testing you're looking for? I'd be interested in taking on this issue

mrocklin:
Good question.  I suspect that Scikit-Learn does this with [their example gallery](http://scikit-learn.org/stable/auto_examples/index.html).  I suspect that @TomAugspurger might know more as well.

mrocklin:
> I'd be interested in taking on this issue

This would be great!

TomAugspurger:
Scikit-Learn uses sphinx-gallery to to convert an `.py` file to
restructured text & jupyter notebooks. I presume they're just running the
python file.

Since we're starting with notebooks, I think the best thing to do is to run
`jupyter nbconvert` on each notebook, to "convert" each to a notebook,
while executing it.

https://nbconvert.readthedocs.io/en/latest/execute_api.html

The main thing is to set a higher timeout I think.

I think the main thing we care about is whether or not the notebook runs.
We don't care about the exact output at all. If the notebook has an
exception, then we want the script executing the notebook to fail, so that
the CI fails.

PR #4, opened 2018-03-22
ML example

TomAugspurger:
Do you think I should add more examples here, or keep this notebook specific to distributed scikit-learn, and add a second notebook for other things from dask-ml?

I'm leaning towards multiple notebooks.


TomAugspurger:
Ah, I just saw #2 :) In that case, maybe I'll expand this to include other examples, and write a followup notebook with more in-depth things like nested CV.

mrocklin:
Thanks for doing this.  I suspect that these examples will be some of the most often clicked :)

machine-learning.ipynb
```diff
@@ -0,0 +1,198 @@
...
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Distributed Training\n",
+    "\n",
+    "<img src=\"images/scikit-learn-logo-notext.png\"/> <img src=\"images/joblib_logo.svg\" width=\"20%\"/> \n",
+    "\n",
+    "Scikit-learn uses [joblib](http://joblib.readthedocs.io/) for single-machine parallelism. This lets you train most estimators (anything that accepts an `n_jobs` parameter) using call the cores of your laptop or workstation.\n",
```
mrocklin: Typo call -> all

mrocklin:
Does this work today?  Are there version requirements?  I'm not getting anything on my dashboard.

Also, should we end with a tiny bit of prediction, just to show that things worked?

TomAugspurger:
Ah, I picked a bad example. It hardcodes the threading backend.

mrocklin:
Is my understanding correct that this is the sort of thing that will be
resolved in the next release?

TomAugspurger:
Yes (maybe). I haven't looked yet whether `LassoCV` needs shared memory, or just prefers it.

TomAugspurger:
Added an example using an algorithm implemented in dask-ml.

mrocklin:
This looks really nice.  I like the combination of using joblib and then using a dask-ml algorithm.  

How do you feel about making the computations somewhat longer when executing?  When running this feels a bit more satisfying.  On my laptop the following changes made me happy:

```python
param_grid = {"C": [0.001, 0.01, 0.1, 0.5, 1.0, 2.0, 5.0, 10.0],
              "kernel": ['rbf', 'poly', 'sigmoid'],
              "shrinking": [True, False]}


X, y = dask_ml.datasets.make_blobs(n_samples=10000000, chunks=1000000, random_state=0, centers=3)


ax.scatter(X[::10000, 0], X[::10000, 1], marker='.', c=km.labels_[::10000],
           cmap='viridis', alpha=0.25);
```

This stayed well under 2GB, and still ran in around 10s or so.  It might be a bit slower on a binder pod though.

TomAugspurger:
Makes sense. We can scale things back if needed.

mrocklin:
This looks great.  Thank you for spending the time on this @TomAugspurger !

TomAugspurger:
👍 everything finished fine on the binder servers. They took about 8s and 12s.

The only minor issue is creating the large dask array `X` caused a warning from distributed

```
distributed.core - WARNING - Event loop was unresponsive in Scheduler for 1.29s.  This is often caused by long-running GIL-holding functions or moving large chunks of data. This can cause timeouts and instability.
distributed.core - WARNING - Event loop was unresponsive in Worker for 1.31s.  This is often caused by long-running GIL-holding functions or moving large chunks of data. This can cause timeouts and instability.
```

mrocklin:
Those have been causing some consternation recently.  I've bumped the limit up to 3s in https://github.com/dask/distributed/pull/1854

PR #6, opened 2018-04-05
Add CI to build notebooks

### LINKED ISSUE #2: Continuous Integration

mholtzscher:
Travis configuration and script to execute all notebooks in repo.

Fixes #2


mholtzscher:
Example Build: https://travis-ci.org/mholtzscher/dask-examples

tests/nbrun.py
```diff
@@ -0,0 +1,35 @@
...
+def run_notebook(in_filepath):
+    if not in_filepath.is_file():
+        raise IOError('File "%s" not found.' % in_filepath)
+    in_filepath = in_filepath.resolve()
+
+    run_path = str(in_filepath.parent)
+
+    nb = nbformat.read(in_filepath.open(), as_version=4)
+    ep = ExecutePreprocessor(
+        timeout=3600, kernel_name='python%d' % sys.version_info[0])
```
mrocklin: The timeout here is in seconds?  If so then I suggest that we reduce this a bit to something like 300 (five minutes)

requirements.txt
```diff
@@ -0,0 +1,14 @@
...
+pandas
+scikit-learn
+matplotlib
+h5py
+graphviz
+nbconvert
+nbformat
+pathlib
+jupyter_client
+ipykernel
```
mrocklin: We currently keep an environment.yml file that we use with the Binder deployment.  I wonder if we can reuse that here instead?  I worry a bit about having to keep the two in sync.  

I tend to use miniconda for testing.  I'm not sure if we want to do that here, but if so I recommend stealing the following lines: https://github.com/mrocklin/streamz/blob/c180551ff5a7eb7aa30bf56242554144f726d45a/.travis.yml#L17-L22

and then following up with 

```
conda env create -f environment.yaml -n test
source activate test
```

tests/nbrun.py
```diff
@@ -0,0 +1,35 @@
...
+        msg = 'Error executing the notebook "%s".\n\n' % in_filepath
+        print(msg)
+        raise
+
+
+print("Running Notebooks...")
+
+for filename in glob.glob('../**/*.ipynb', recursive=True):
+    print(filename)
+    run_notebook(Path(filename))
```
mrocklin: I'm curious, was this code inspired by another project's solution or is it from scratch?

tests/nbrun.py
mholtzscher: Yes, it is in seconds. I meant to change it but forgot. I agree with 300.

requirements.txt
mholtzscher: I will investigate using miniconda instead

tests/nbrun.py
mholtzscher: I looked at a couple other repos and combined what I liked from each

.travis.yml
```diff
@@ -0,0 +1,17 @@
+language: python
+
+python:
+    - "3.5"
+    - "3.6"
```
mholtzscher: @mrocklin  what python versions do you want me to target?

mrocklin: Perhaps controversially, I suggest that we only do 3.6 for the examples

TomAugspurger:
Thanks @mholtzscher.

Would you mind (temporarily) adding an error in a notebook and ensure that the failure bubbles all the way up to fail the build? Everything looks like it should, but just want to verify.

mholtzscher:
@TomAugspurger Here you go! https://travis-ci.org/mholtzscher/dask-examples/builds/362830883

TomAugspurger:
Great, thanks! Happy to see the failure :)

mrocklin:
This looks good to me.  I've taken the liberty of making a small change and moving the extra dependencies to the travis.yml file rather than the environment.yml file, which is also shared with the binder deployment, which we'd like to keep lean if possible.

mrocklin:
Merged.  Thanks @mholtzscher !

PR #14, opened 2018-06-29
ENH: add xgboost example

stsievert:
Implement an xgboost example.

Mirror the example at https://dask-ml.readthedocs.io/en/latest/examples/xgboost.html. That example downloads data and does some preprocessing, which takes a non-trivial amount of time (at least a minute or two). Plus, the ROC curve doesn't look to good.

This example creates synthetic data, which is created fast (though, it does point to other example and the video @mrocklin recorded). This example is visible at https://gist.github.com/stsievert/b797b4e4d510a4e417a8db33eb6601c6

Creation of synthetic data allows  to show the comparison with the ground truth: turns out XGBoost does a good job at identifying which features are important:

Plus the ROC curve looks better:


xgboost.ipynb
```diff
@@ -0,0 +1,227 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "This example is similar to the example at https://dask-ml.readthedocs.io/en/latest/examples/xgboost.html, but this example takes a while to run because it uses a real dataset. There is a recorded screencast that walks through this example  at https://www.youtube.com/watch?v=Cc4E-PdDSro, with comments from the lead Dask developer, Matthew Rocklin.\n",
```
mrocklin: I recommend that we start strong with the topic at hand, rather than discuss the other examples that are out there.  Maybe something like "Dask and XGBoost can work together to train data with gradient boosted trees.  This notebook shows how"

The context of other examples that may exist may not interest new users.  It's usually an accurate guess that for anyone looking at an example it is their first time seeing the topic.  If we do want to provide a link to the other example, lets do it at the bottom in a "Learn more" section

xgboost.ipynb
```diff
@@ -0,0 +1,227 @@
...
+    "\n",
+    "This example is similar but uses synthetic data instead, to allow it to be easily runnable. This example uses a Dask wrapper around XGBoost.\n",
```
mrocklin: Lets also not mention me, or any other developer by name.  If we want to include a screencast of this notebook we should just say "Here is a screencast of this notebook".

xgboost.ipynb
```diff
@@ -0,0 +1,227 @@
...
+    "\n",
+    "<img src=\"http://dask.readthedocs.io/en/latest/_images/dask_horizontal.svg\" width=\"30%\" alt=\"Dask logo\"> <img src=\"https://raw.githubusercontent.com/dmlc/dmlc.github.io/master/img/logo-m/xgboost.png\" width=\"25%\" alt=\"Dask logo\">"
```
mrocklin: Thanks for the images.  Can I also ask for a header title?  Perhaps something like

```markdown
Dask and XGBoost
===============
```

xgboost.ipynb
```diff
@@ -0,0 +1,227 @@
...
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "import os\n",
+    "from dask import compute, persist\n",
+    "from dask.distributed import Client, progress\n",
+    "\n",
+    "client = Client(os.environ.get(\"DISTRIBUTED_ADDRESS\"))  # connect to cluster \n",
```
mrocklin: You'll want to create a client here normally.  I recommend looking at the other example notebooks in this repository to see how.

I'm curious, where did you find this line of code?  Were you working from an existing example somewhere?  (if so, I need to scrub that example)

xgboost.ipynb
```diff
@@ -0,0 +1,227 @@
...
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "import dask.array as da\n",
+    "from dask_ml.model_selection import train_test_split\n",
+    "import numpy as np\n",
+    "\n",
+    "n, d = int(100e3), 20\n",
```
mrocklin: Rather than `int(100e3)` can I ask for `1e5` or `100000` (preferred)

xgboost.ipynb
```diff
@@ -0,0 +1,227 @@
...
+    "da.random.seed(42)\n",
```
mrocklin: Is this line necessary for smooth operation?  If not then I recommend removing it.  In general the less that a novice user can get tripped up by the better. 

I'm curious, as a graduate student have you ever had to teach introductory computer science?  Have you had the pleasure of making course materials that seemed perfectly clear only to have hundreds of students get entirely confused by every other line?  If not, I recommend it.  It's a tremendous learning opportunity :)

xgboost.ipynb
```diff
@@ -0,0 +1,227 @@
...
+    "\n",
+    "client = Client(os.environ.get(\"DISTRIBUTED_ADDRESS\"))  # connect to cluster \n",
+    "client"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "First we create a bunch of synthetic data:"
```
mrocklin: Can I ask for small section headers to break up the flow of the notebook?  I recommend looking at existing notebooks to get a sense of style.  For example, this might have a section title like 

```
### Make an artificial dataset
```

xgboost.ipynb
```diff
@@ -0,0 +1,227 @@
...
+    "import xgboost as xgb"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "dask-xgboost is a small wrapper around xgboost, and will behave the same as xgboost.\n",
+    "\n",
+    "During training it will take care of handling the data, and allow xgboost to use it's own distributed scheduler."
```
mrocklin: The first "it" is ambiguous here between Dask and XGBoost.  Lets use "Dask" to be explicit.  Lets also expand "handling" to "loading, cleaning, and pre-processing" or something similar.  Lets capitalize XGBoost to be polite to other software projects.  I recommend that we remove the term "scheduler", which is not commonly known by novices, and perhaps use something like "and leverage XGBoost's own distributed training system" instead.

We're not targetting people well versed in machine learning or distributed systems.  We're targetting readers who are far less sophisticated in this realm, but quite sophisticated in their own domain, like genomics or insurance.

xgboost.ipynb
```diff
@@ -0,0 +1,227 @@
...
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "%%time\n",
+    "params = {'objective': 'binary:logistic', 'nround': 1000, \n",
+    "          'max_depth': 5, 'eta': 0.01, 'subsample': 0.5, \n",
+    "          'min_child_weight': 0.5}\n",
+    "\n",
+    "bst = dxgb.train(client, params, X_train, y_train)\n",
```
mrocklin: It would be good to remove the need to pass in `client` here in dask-xgboost and instead use `dask.distributed.default_client` if client is not explicitly provided.  Can I ask you to submit a small PR to the dask-xgboost library with this fix?  

I'm not sure about the best way to deprecate this cleanly.  It might also be a good idea to open an issue to have a longer conversation with @TomAugspurger .

xgboost.ipynb
```diff
@@ -0,0 +1,227 @@
...
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "import matplotlib.pyplot as plt\n",
+    "import pandas as pd\n",
+    "\n",
+    "if isinstance(w_star, da.Array):\n",
+    "    w_star = w_star.compute()\n",
```
mrocklin: We know that it is a dask array here, right?  My guess is that you want to make this cell so that you can run it many times.  If so I might suggest one of two approaches:

1.  Use a different variable name for the dask array and numpy version
2.  Use `w_star = dask.compute(w_star)[0]`, which will work in both cases

xgboost.ipynb
```diff
@@ -0,0 +1,227 @@
...
+    "\n",
+    "bst = dxgb.train(client, params, X_train, y_train)\n",
+    "bst"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "The `bst` object is a regular `xboost.Booster` object, and has all the familar methods available. For example,"
```
mrocklin: Trails off with "For example,".  Not sure if this was intentional or not.

xgboost.ipynb
```diff
@@ -0,0 +1,227 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "y_hat = dxgb.predict(client, bst, X_test)\n",
+    "y_hat = client.persist(y_hat)\n",
```
mrocklin: I recommend that we avoid the `client.persist` method in all cases, instead preferring the `.persist` method of the series object, or the `dask.persist` function.  In this case I would do the following:

```python
y_hat = dxgb.predict(client, bst, X_test).persist()
```

mrocklin: We might also want to call out that we can use `dask_xgboost` to predict in parallel in prose.

xgboost.ipynb
```diff
@@ -0,0 +1,227 @@
...
+    "df.plot.bar(ax=axs[1])\n",
+    "axs[1].set_title('Ground truth feature importance')\n",
+    "plt.show()\n"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Or we can plot the ROC curve after we do some predictions:"
```
mrocklin: As someone who is a novice in machine learning I have to admit that I don't have a strong understanding of ROC curves.  It might be good to spell out the acronym once in the style of `Three Letter Acronym (TLA)` and then state that "we see that results are good because the curve tends towards the upper left", just to give the novice reader the ability to appreciate how well things worked out.

xgboost.ipynb
```diff
@@ -0,0 +1,227 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "import dask_xgboost as dxgb\n",
+    "import xgboost as xgb"
```
mrocklin: Thoughts on leaving these as full names like `xgboost` and `dask_xgboost`?  I'm not sure how common these pronouns are.

xgboost.ipynb
stsievert: I was wondering about it but it worked so I kept going. I copied and pasted from https://dask-ml.readthedocs.io/en/latest/examples/xgboost.html.

xgboost.ipynb
stsievert: >  My guess is that you want to make this cell so that you can run it many times.

Exactly. I was iterating on this cell.

xgboost.ipynb
stsievert: I've expanded, and have included a paragraph about what XGBoost is and why it's useful, and it provides good motivation for the rest of the notebook. I left same thing for Dask out because if they're looking at this example they already want to learn about Dask. Let me know if you'd like me to change that.

xgboost.ipynb
stsievert: They're easier to type and easier to understand. I don't see a downside. I've changed them.

xgboost.ipynb
stsievert: I've added an explanation for what ROC curves are.

stsievert: And a link too.

xgboost.ipynb
stsievert: Eek, I didn't read closely enough – it's in the other example too. That should be modified too.

xgboost.ipynb
stsievert: I've added

> XGBoost will leverage their own distributed training system using all the workers that Dask has available. Dask sets XGBoost up, gives XGBoost data and lets XGBoost do it's training in the background using all the workers Dask has available.

Does this clearly express what dask-xgboost does? I adopted the last line from [your blogpost on XGBoost][1]. Do I have your permission to use it here?

[1]:http://matthewrocklin.com/blog/work/2017/03/28/dask-xgboost

xgboost.ipynb
stsievert: This is important for some of the results I mention – I specifically call out some of the labels in the plot as examples.

I haven't had that experience yet, but I did have the fortune of giving a paper presentation for my quals. Believe me, giving a paper presentation to both a world-class expert on that particular paper and a layman is also a learning opportunity.

xgboost.ipynb
stsievert: Yup, implemented before I saw this. It'll give more structure, and help people who are skimming.

stsievert:
I see most of the other notebooks in this repo are without their outputs in the notebook. Letting the outbooks be coded in the notebook would allow users to view the notebooks on GitHub alone, and try binder if they want.

mrocklin:
I've pushed a few minor edits.  This is good to go from my perspective.  @TomAugspurger any final thoughts?

mrocklin:
> I see most of the other notebooks in this repo are without their outputs in the notebook. Letting the outbooks be coded in the notebook would lower the barrier to viewing to viewing these notebooks by allowing users to try binder if they want. Why don't we include the outputs in the notebooks?

Moved to https://github.com/dask/dask-examples/issues/16

binder/environment.yml
```diff
@@ -13,3 +13,5 @@ dependencies:
   - h5py
   - pip:
     - graphviz
+    - dask_xgboost
+    - xgboost
```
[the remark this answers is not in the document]
mrocklin: Ping @stsievert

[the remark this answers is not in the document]
stsievert: Thanks for the ping @mrocklin.

mrocklin:
@stsievert I made a few edits to this notebook similar to what we just went over.  Would you might looking things over and seeing if you're ok with them?

stsievert:
I've simplified this a lot, by both by reducing unnecessary prose and simplifying the code.

I've pushed with the noteboko with outputs, but can remove them before merge.

mrocklin:
Seems good to me!

mrocklin:
This is in.  Thanks @stsievert !

PR #15, opened 2018-06-29
ENH: add an example with dask-ml's Incremental

stsievert:
I'd appreciate a review of this. I tried to point out the features and limitations of `Incremental`, and walked through a basic training + testing example. I included a note about https://github.com/dask/dask-ml/pull/247, and how Incremental is responsible for data management and sklearn is responsible for the computation on the data.

A rendered notebook is available at https://gist.github.com/stsievert/fbd2da97924e827bdb2f015aff7fe9b1

cc @TomAugspurger @mrocklin


machine-learning-scikit-learn.ipynb
```diff
@@ -0,0 +1,333 @@
...
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "%%time\n",
+    "# takes about 1min\n",
+    "\n",
+    "from distributed.metrics import time\n",
+    "data = []\n",
+    "start = time()\n",
+    "for iteration in range(10):\n",
```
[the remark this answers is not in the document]
TomAugspurger: It would also be good to note (somewhere) that these multiple epochs + distributed provide an advantage over writing your own for loop to load in chunks of data off disk: With distributed you only pay the IO cost once.

machine-learning-scikit-learn.ipynb
```diff
@@ -0,0 +1,333 @@
...
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "import pandas as pd\n",
+    "import matplotlib.pyplot as plt\n",
+    "df = pd.DataFrame(data)\n",
+    "df.plot(x='epochs', y='score')\n",
+    "plt.title('Score over iterations')\n",
+    "plt.ylabel('Score')\n",
```
TomAugspurger: "Trainining Score"

machine-learning-scikit-learn.ipynb
```diff
@@ -0,0 +1,333 @@
...
+    "Let's predict, than compute the accuracy"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "y_hat = inc.predict(X_test)\n",
```
TomAugspurger: I think scikit-learn uses `y_pred`, so people may be more familiar with that (though I'm partial to `y_hat`, in this case it's best to go where the people are).

machine-learning-scikit-learn.ipynb
```diff
@@ -0,0 +1,333 @@
...
+    "errors = np.abs(y_hat - y_test) / 2   # divide by 2 to get 0-1 labels\n",
```
TomAugspurger: `dask_ml.metrics.accuracy_score`

machine-learning-scikit-learn.ipynb
```diff
@@ -0,0 +1,333 @@
...
+   "metadata": {},
+   "source": [
+    "## Grid search"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "This accuracy might be the best possible, but there a lot of parameters to feed `SGDClassifier` that can drastically change the performance. We'll consider two of the most basic parameters, specifying a different loss function and different amount of penalty. This is the problem of hyperparameter optimization, something Dask-ML supports: https://dask-ml.readthedocs.io/en/latest/hyper-parameter-search.html\n",
```
TomAugspurger: Is the first phrase missing a "not"? Somethying like

"This accuracy might not be the best possible; there are a lot of parameters..."

machine-learning-scikit-learn.ipynb
```diff
@@ -0,0 +1,333 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "from dask_ml.model_selection import GridSearchCV\n",
+    "params = {'alpha': np.logspace(-6, 0, num=10),\n",
```
TomAugspurger: I would write this like:

```python
# small parameter grid for the example
params = {
    'alpha': [0.1, 10],
    'loss': ['hingle', 'log'],
}

# An example of a larger parameter grid.
# params = {
#    'alpha': np.logspace(-6, 0, num=10),
#    'loss': ['hinge', 'log', 'modified_huber', 'squared_hinge', 'perceptron']
# }
```

machine-learning-scikit-learn.ipynb
```diff
@@ -0,0 +1,333 @@
...
+    "But what if other losses were equally good? Let's see a visualization of the different losses *and* different regularization parameters."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "import seaborn as sns\n",
```
TomAugspurger: I slightly prefer 

```python
g = sns.factorplot(x='param_alpha', y='mean_test_score', data=df,
                   aspect=3,
                   hue='param_loss')
```

for visualizing these.

gh


ticklabels would need to be cleaned up.

will also need to ensure seaborn is available on the binder worker.

machine-learning-scikit-learn.ipynb
```diff
@@ -0,0 +1,333 @@
...
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "import sklearn.datasets\n",
+    "import dask.array as da\n",
+    "import numpy as np\n",
+    "n, d = int(100e3), 100\n",
+    "\n",
+    "X = da.random.normal(size=(n, d), chunks=n // 10)\n",
```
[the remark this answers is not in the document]
stsievert: > dask/dask-ml#255 would cover this right?

It would help. I've gotten rid of `w_star` – I think I was using it for ground truth, but I don't actually use it later.

machine-learning-scikit-learn.ipynb
```diff
@@ -0,0 +1,333 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "%%time\n",
+    "# takes about 1min\n",
```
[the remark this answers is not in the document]
stsievert: I've reduced the number of iterations so it takes 1min with 2 workers on my machine.

machine-learning-scikit-learn.ipynb
stsievert: I was using raw dask arrays to provide some basics on how to use them. I've tried to clarify that, and re-ordered too.

machine-learning-scikit-learn.ipynb
stsievert: I greatly prefer this instead of the heatmap.

TomAugspurger:
Hmm I had a review pending, but apparently lost it.

Overall, the content here is solid. We may want to wait on the resolution to https://github.com/dask/dask-ml/pull/258, as that'll be an API breaking change.

stsievert:
Could I have another review on this PR? I'm satisfied with the prose and the performance (which has spawned a couple other issues and PRs)

stsievert:
I've pushed the notebook with outputs. I can strip them before removal depending on #16.

machine-learning-scikit-learn.ipynb
```diff
@@ -0,0 +1,744 @@
...
+   "metadata": {},
+   "source": [
+    "# Scikit-learn + Dask"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Scikit-learn is a popular machine learning library whose API is easy to use because it has been inspired by other machine learning libraries. It encompasses a wide variety of problem formulations, ranging from simple linear models to more complex clustering or decomposition formulations. Scikit-learn works great for problems that fit on a single machine, but struggles for larger problems. This has even gotten a page in their user guide, \"[Strategies to scale computationally: bigger data][strat]\".\n",
```
TomAugspurger: I'm not sure about the origins of scikit-learn's design, but I often see *it* cited as the inspiration for other APIs (e.g. https://spark.apache.org/docs/latest/ml-pipeline.html#main-concepts-in-pipelines)

mrocklin:
Some comments, mostly language related.  My apologies for not going in-line:

-  Avoid words like *plethora* that may not be known to non-native english speakers
-  Similarly with phrases like "ever hear of teaching to the test?"  This can be confusing if you don't know what it means, which is possibly most of our readers.  This then diminishes leads you into a tangent of trying to figure out if teaching is another term for training/fitting, etc..
- More generally, we should probably avoid jokes and cleverness in all prose except perhaps blogposts.  If you were to reduce the amount of prose in this notebook by a factor of two or three I think that that would be for the best.  A user probably wants to move through this in a couple minutes.  We should endeavor to give them the minimum amount of information that they need to move on.  They will appreciate that we respect their time.
-  Remove phrases like "as usual" and "of course" which assume that the user knows what's going on.  A wide variety of users will read these examples, and for many of them all of this will be new.
-  I recommend removing the manual computation of accuracy.  I get that it can be useful to show them dask.arrays, but they can also look at the dask array examples for this if that's their interest.  Lets focus in on the use of Incremental
-  I also recommend removing the GridSearch section for the same reason.  Lets get them in and out quickly.  
- Calling `foo.persist(); bar.persist()` is sort of an anti-pattern.  It's nicer to call them all at once with the dask.persist function.  

As an example, here is what I recommend for how to handle persist and da.unique.  I recommend the following cells directly after creating the training data.

```
### Persist data in memory

This dataset is small enough to fit in distributed memory, so we call `dask.persist` now.

    X_train, y_train, X_test, y_test = dask.persist(X_train, y_train, X_test, y_test)

If you are working in a situation where your dataset does not fit in memory then you should skip this step.  Everything will still work, but will be slower and use less memory.

### Pre-compute classes

We pre-compute the classes from our training data

    classes = da.unique(y_train).compute()
    classes

...

### Create Scikit-Learn Estimator
 
Your explanation looks good (other than "as usual")

    sgd = ...

### Wrap with Dask-ML Incremental 

Incremental is a meta-estimator that ...
```

I've found that this level of prose/code is easily digestible by peole with short attention spans (most readers).  Separating things out by section headers with brief informative names helps them to understand the scope of the step of the computation that they are about to learn about.

machine-learning-scikit-learn.ipynb
```diff
@@ -0,0 +1,744 @@
...
+   "metadata": {},
+   "source": [
+    "## Dask setup"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Creating a distributed scheduler will provide good feedback because we can view the dask.distributed dashboard. This will provide progress and performance metrics, including visualization of jobs that are running."
```
TomAugspurger: I think this is redundant with the previous.

machine-learning-scikit-learn.ipynb
```diff
@@ -0,0 +1,744 @@
...
+   "source": [
+    "classes = da.unique(y).compute()\n",
+    "inc.fit(X_train, y_train, classes=classes)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Now, let's call `partial_fit` multiple times.\n",
```
TomAugspurger: I would remove this block, since implementing `max_iter` for Incremental will I think make it obsolete?

machine-learning-scikit-learn.ipynb
```diff
@@ -0,0 +1,744 @@
...
+   "metadata": {},
+   "source": [
+    "## Grid search"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "There a lot of parameters to feed `SGDClassifier`, and this accuracy depends on those parameters, and it may not be as high as it can be. We'll consider two of the most basic parameters, specifying a different loss function and different amount of penalty. This is the problem of hyperparameter optimization, something Dask-ML supports: https://dask-ml.readthedocs.io/en/latest/hyper-parameter-search.html\n",
```
TomAugspurger: Typo "There a".

machine-learning-scikit-learn.ipynb
stsievert: I want to leave this block we're show casing `Incremental`, which will very likely be used for other tools to get some idea or progress/etc. I would imagine that the `data` dictionary I use will be used, and probably with more data.

I have restructered a bit, to move the persisting data section below the "calling `partial_fit`" header.

stsievert:
I have addressed both @mrocklin and @TomAugspurger's comments. I've pushed the notebook with the outputs again.

> I also recommend removing the GridSearch section for the same reason. Lets get them in and out quickly.

Good idea. I've removed the model selection section. I have briefly mentioned this alongside the Dask joblib integration in a "Learn more" section.

TomAugspurger:
FYI I’m traveling today and will likely be busy on pandas things the next few days, so don’t let me hold any of these up.

________________________________

mrocklin:
I'm taking a look through this now. I may make a few edits directly if that's ok.

stsievert:
> I may make a few edits directly if that's ok.

Thank you, it's appreciated. That's more than okay.

machine-learning/incremental.ipynb
```diff
@@ -0,0 +1,542 @@
...
+    "Many Scikit-Learn estimators implement a `partial_fit` method to enable incremental learning in batches.  Typically it is up to the user to collect batches of data to give to a Scikit-Learn estimator, which then updates a single estimator.\n",
+    "\n",
+    "```python\n",
+    "est = SGDClassifier()\n",
+    "\n",
+    "for filename in filenames:\n",
+    "    df = pandas.read_csv(filename)\n",
+    "    x, y = ...  # get data from CSV using the Pandas API\n",
+    "    est.partial_fit(x, y)\n",
+    "```\n",
```
stsievert: Do you have any examples for "this is a common operation"? The first thing I would do is shuffle the dataset before calling the optimizer on each block.

mrocklin: > The biggest edits I made were adding some motivation for calling dask.persist.

Sounds fine to me.  As a general note, there are always going to be lots of things that we'll do in any of these examples that are important that we won't be able to tell them about.  We'll necessarily skip a lot of potential wisdom we could deliver in order to focus on the topic at hand.

In this case though I agree that including some information about persist is good, especially because it's nice and concise.

mrocklin:
This is in.  Thanks @stsievert !

PR #18, opened 2018-07-06
MAINT: make Incremental example quick on binder

stsievert:
I played with this on binder, and it was slow. I've made some modifications to make it faster and to have the user wait less time. These modifications are

* decreasing number of chunks by factor of 10
* decreasing number of examples by factor of 10


mrocklin:
Thanks for following up and making sure things were smooth.  Hopefully this will also improve a bit when we release and drop the tokenize costs.

Issue #16, opened 2018-07-02
Include outputs in notebooks


mrocklin:
In https://github.com/dask/dask-examples/pull/14#issuecomment-401507027 @stsievert says 

> I see most of the other notebooks in this repo are without their outputs in the notebook. Letting the outbooks be coded in the notebook would lower the barrier to viewing to viewing these notebooks by allowing users to try binder if they want. Why don't we include the outputs in the notebooks?

This seemed to be worth discussion, so I've raised a separate issue for it here.

mrocklin:
When working in project repositories like dask/dask or dask/dask-ml we always strip outputs of example notebooks to keep the size of the repository small.  Otherwise every time someone re-runs the notebook and produces an output plot the repo grows by a several hundred kB.  Aggregated over years of time the repository bloats out.

However, hygiene around this repository isn't as strict as with those.  Modest outputs might be ok?  What do others think?

martindurant:
I agree that "examples" have a decent argument for being rendered, where as "tutorials" should not be, since they are intended to be downloaded and run by users.

mrocklin:
I'm going to close this.  Our solution was to run the notebooks and then place their evaluated forms on https://example.dask.org

stsievert:
> then place their evaluated forms on https://examples.dask.org

:+1: that’s an even better solution!

PR #19, opened 2018-07-06
Switch delayed notebook to use processes

mrocklin:
This started failing after the change to 1.22.1
I don't yet know why

### DIFF
--- a/delayed.ipynb
+++ b/delayed.ipynb
@@ -40,7 +40,7 @@
    "outputs": [],
    "source": [
     "from dask.distributed import Client, progress\n",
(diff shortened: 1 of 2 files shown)


martindurant:
Did we not want to leave results in these notebooks, so they would render for those not running locally?

mrocklin:
I think that the current convention is not to.  We should discuss that here: https://github.com/dask/dask-examples/issues/16 .  For now I'm going to merge this on green light.  It fails live.

These outputs had actual errors in them, so clearing them is probably for the best.

PR #24, opened 2018-07-18
Add Dataframe notebooks

mrocklin:
This starts to add a series of notebooks for dask dataframe, similar to what we have now for machine learning.  

I've also included a screencast in the first notebook as an experiment.  If we switch to JupyterLab we could also have this running on the side (if the users right clicks and moves the output to a new cell)

Binder running here: https://mybinder.org/v2/gh/mrocklin/dask-examples/dataframes?filepath=dataframes%2F01-data-access.ipynb

I'm happy to continue going here, but I'd like to get some feedback first if possible, especially from @AlbertDeFusco if he has time.

### DIFF
--- a/dataframes/01-data-access.ipynb
+++ b/dataframes/01-data-access.ipynb
@@ -10,6 +10,31 @@
    ]
   },
   {
+   "cell_type": "code",
+   "execution_count": 1,
+   "metadata": {},
+   "outputs": [
+    {
+     "data": {
+      "text/html": [
+       "<iframe width=\"560\" height=\"315\" src=\"https://www.youtube.com/embed/0eEsIA0O1iE?rel=0&amp;controls=0&amp;showinfo=0\" frameborder=\"0\" allowfullscreen></iframe>"
+      ],
+      "text/plain": [
+       "<IPython.core.display.HTML object>"
+      ]
+     },
+     "execution_count": 1,
+     "metadata": {},
+     "output_type": "execute_result"
+    }
+   ],
+   "source": [
+    "from IPython.display import HTML\n",
+    "\n",
+    "HTML('<iframe width=\"560\" height=\"315\" src=\"https://www.youtube.com/embed/0eEsIA0O1iE?rel=0&amp;controls=0&amp;showinfo=0\" frameborder=\"0\" allowfullscreen></iframe>')"
+   ]
+  },
+  {
    "cell_type": "markdown",
    "metadata": {},
    "source": [
@@ -121,7 +146,19 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "We can read many of these files with the dask.dataframe.read_csv function"
+    "We can read one file with `pandas.read_csv` or many files with `dask.dataframe.read_csv`"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "import pandas as pd\n",
+    "\n",
+    "df = pd.read_csv('data/2000-01-01.csv')\n",
+    "df.head()"
    ]
   },
   {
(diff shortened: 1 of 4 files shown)


AlbertDeFusco:
Here's a few things:

## 01-data-access
* perhaps mention that Dask can handle many files that in aggregate store more data than available memory
* Dask can read globs of CSV files that have the same column names without loops or appends. The example shown here is that data is stored daily in separate CSV files.

>Whenever we operate on our dataframe we read through all of our CSV data so that we don't fill up RAM. This is very efficient for memory use, but reading through all of the CSV files every time can be slow.

I'd recommend

>Whenever we operate on our dataframe only a small chunk of the data is read by each processor at a time. This means we won't fill up our RAM when operating on a large collection of data. Even though this is very memory efficient CSV files can be slow in general, which is also true for Pandas.


## 02-groupby

I hadn't seen `split_out` before. Would the best practice be that `split_out` should be just small enough so that a single partition can fit into memory?

Custom aggregations be only be used with `.apply()`, not `.agg()`, like this function below? Perhaps *groupby-aggregation* could refer to the [common aggregation functions](https://pandas.pydata.org/pandas-docs/stable/groupby.html#aggregation)

```python
def count_outliers(group):
    z = (group - group.mean()) / group.std()
    return (z.abs() >= 1.8).sum()

df.groupby('CAT')['VALUE'].apply(count_outliers, meta='int64')
```

AlbertDeFusco:
Would it be possible to create the CSV files during the binder initialization? Or maybe just as a single function imported from a py file?

```python
import os
import datetime

if not os.path.exists('data'):
    os.mkdir('data')

def name(i):
    """ Provide date for filename given index
    
    Examples
    --------
    >>> name(0)
    '2000-01-01'
    >>> name(10)
    '2000-01-11'
    """
    return str(datetime.date(2000, 1, 1) + i * datetime.timedelta(days=1))

def make_csv_files(size='small'): 
    start = '2000-01-01'
   
    if size == 'medium':
        end='2000-12-31'
    elif size == 'large':
        end='2010-12-31'
    else:
        end='2000-01-31'

    df = dask.datasets.timeseries(start, end)
    df.to_csv('data/*.csv', name_function=name);
```

mrocklin:
Thanks Albert.  Also, do you have any thoughts on the approach of including
a video?  Is the style of the video appropriate?  Should it be more
detailed, etc..?

AlbertDeFusco:
The videos are nice. They are concise showing what the user should see as they work with the same code along with useful background information.

dataframes/01-data-access.ipynb
```diff
@@ -0,0 +1,365 @@
...
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "import os\n",
+    "import datetime\n",
+    "\n",
+    "if not os.path.exists('data'):\n",
+    "    os.mkdir('data')\n",
```
TomAugspurger: For Py3+ (if we're OK with that) there's `os.makedirs('data', exist_ok=True)`, rather than the if block.

dataframes/02-groupby.ipynb
```diff
@@ -0,0 +1,226 @@
...
+    "df"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "This dataset is small enough to fit in memory, so we persist it now.\n",
+    "\n",
+    "You would skip this step if your dataset becomes too large to fit into memory."
```
TomAugspurger: I get the feeling these are meant as introductory notebooks. Should we clarify which "memory" may be too small at this point? I could see confusion on whether this is the memory of the client or the cluster.

TomAugspurger: But, if this is introductory, that may be too confusing at this point.

dataframes/01-data-access.ipynb
mrocklin: I think I'll pass on this for now just out of laziness.  The current solution also seems easy to understand to me.

dataframes/02-groupby.ipynb
mrocklin: I've added the word "cluster's" before memory here to disambiguate

mrocklin:
@TomAugspurger ok to merge?

TomAugspurger:
Yep!

PR #25, opened 2018-07-21
Made an as simple as possible example of calling R code from a Dask c…

tsktsktsk123:
Made an as simple as possible example of calling R code from a Dask cluster to show that it can work. Love to hear some feedback!

### DIFF
--- a/dask-rpy2-R.ipynb
+++ b/dask-rpy2-R.ipynb
@@ -1,0 +1,185 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# **Executing R code distributed through Dask**\n",
+    "\n",
+    "Dask can be used to distribute the executing of R code across a cluster.\n",
+    "In this example each worker gets the same R code to execute on different data.\n",
+    "It does not parallize the R algorithms or any of the code. The R code runs isolated in the thread of their worker.\n",
+    "\n",
+    "This example is an extension of the [Dask Dataframe example](https://github.com/dask/dask-examples/blob/master/dataframe.ipynb).\n",
+    "\n",
+    "### **Requirements for each worker:**\n",
+    "- [rpy2](https://rpy2.bitbucket.io/)\n",
+    "- [R](https://www.digitalocean.com/community/tutorials/how-to-install-r-on-ubuntu-16-04-2) \n",
+    "\n",
+    "\n",
+    "### **Install rpy2**\n",
+    "**NOTE:** Make sure that R is already installed and available else this will not work properly."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "!pip install rpy2"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "#### **Imports**"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 1,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "import pandas as pd\n",
+    "import numpy as np\n",
+    "\n",
+    "import dask.dataframe as dd\n",
+    "from dask.distributed import Client\n",
+    "\n",
+    "import rpy2.robjects as ro\n",
+    "from rpy2.robjects import pandas2ri\n",
+    "from rpy2.robjects.packages import importr"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "#### **Create dummy data**\n",
+    "\n",
+    "We create a random timeseries of data with the following attributes:\n",
+    "\n",
+    "    It stores a record for every 1 hour of the year 2000\n",
+    "    It splits that year by month, keeping every month as a separate Pandas dataframe\n",
+    "    Along with a datetime index it has columns for names, ids, and numeric values\n"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 2,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "df = dd.demo.make_timeseries('2000-01-01', '2000-12-31', freq='1h', partition_freq='1M',\n",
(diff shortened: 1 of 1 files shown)


mrocklin:
Thank you @tsktsktsk123 .  This looks great!  I hope you don't mind a few nitpicky comments, mostly style:

1.  It looks like we're not using the numpy or dask.distributed imports.  We should remove them
2.  Can I ask you to provide spaces around assignment equalities as per PEP8

    ```python
    x=123  # don't
    x = 123  # do
    ```
3.  Can I ask for a single space after the `#` when doing a comment

    ```python
    x = 123  #don't
    x = 123  # do
    ```
4.  For this comment: `Return the output. For more info about extracting see: http://rpy2.readthedocs.io/en/version_2.8.x/vector.html#extracting-python-style` I recommend that we leave the `Return the output` bit but move the rest down to a markdown cell after this code cell.  Generally we should avoid comments that are so long that they wrap around the screen.  Instead long explanations should probably be moved to markdown.  I think that the other comments look great though and balance well between being informative and terse.
6.  The title implies that we're distributing code, but in this case we're just using the local scheduler.  Maybe change the title to `Parallelize R code with Dask` ?

mrocklin:
@TomAugspurger how should we handle the test failure here?  It looks like the testing scripts don't pick up the `pip install` line

tsktsktsk123:
@mrocklin Thanks for the feedback! I worked through your points.

Regarding the pip install line: I added it to be complete with the remaining examples but I must add that I feel it's better leaving it out. The presence of the pip install line implies that it is the only installation required for this example to work, which isn't sadly - R needs to be installed as well and afaik rpy2 won't install correctly without R being present.

Not sure how the tests are made but some options are:

1) Remove the line and leave it to the reader to install the correct requirements
2) Add R to base image used for testing, allowing the tests to run this example in their entirety. 
3) Some other options I can't think of.

Thanks.

mrocklin:
> Add R to base image used for testing, allowing the tests to run this example in their entirety.

So, to be clear this isn't just for testing, this is also for live execution with binder (you can run your own notebook here at https://mybinder.org/v2/gh/tsktsktsk123/dask-examples/master) and we'd like to keep the image somewhat small so that it runs quickly.  I would be curious to know how expensive it is to add r-base.  My guess is that it's not that big, but it would be good to verify.

> Some other options I can't think of.

We could also conda install R.  I think that there is some standard channel of R packages.  Conda installing takes a bit more time though, which is why we generally prefer pip for quick interactive use cases like the binder notebooks.

TomAugspurger:
I share @mrocklin's concerns about startup time. @tsktsktsk123 do you have a rough idea how large `r-base` is?

tsktsktsk123:
Not entirely sure actually. I install it with a playbook role each time and it takes 30s - 1min approx. 

I'll try to clock the options and get back somewhere this week. I'm thinking this saturday but perhaps earlier.

Also this runner took 6min59s and the [regular](https://travis-ci.org/dask/dask-examples/pull_requests) times take between 7 and 8 minutes: It does not seem to take longer even though the tests seemed to pass. Still not entirely sure if it tested the notebook correctly as I'm not seeing an r-base installation in the runner logs. Does the notebook as is work in the binder environment?

tsktsktsk123:
Didn't manage to time the installation as thoroughly as I wanted and I'm going for holidays tomorrow for 3 weeks. Though running ` conda install R` on a clean VPS took about 4-5 minutes (!). Just running `sudo apt-get install r-base` didn't took longer than 1 minute though. Installing rpy2 does take some time as well so I think  installing with `apt` and installing rpy2 will end up taking about 1-2 minutes extra.

Regarding the tests, I've tried to make some edits to have the runner install r but haven't got it working yet. After my holidays I got time to pick this up thoroughly.

mrocklin:
cc @rsignell-usgs

mrocklin:
@dsludwig if you find yourself with free time you might consider finding a nice way to push this along that balances between user experience (long wait times on binder to apt/conda install R) and image size (how expensive is it to add R to the docker image)?

This came up in the pangeo meeting.

rsignell-usgs:
@mrocklin, thanks for the ping.  I was able to successfully run this notebook using a `conda-forge` only environment, with no need to `sudo` anything, and nothing from `defaults` channel !

```
name: rtry
channels:
  - conda-forge
dependencies:
  - dask
  - dask-ml
  - distributed
  - numpy
  - pandas
  - scikit-learn
  - matplotlib
  - nbserverproxy
  - nomkl
  - h5py
  - rpy2
  - xgboost
  - graphviz
  - dask-xgboost
  - seaborn
```
The only stumbling block was realizing that the required package from `conda-forge` is `xgboost`, not `py-xgboost`.  I not sure why anaconda chose to call this package `py-xgboost` in the defaults channel.  I believe `xgboost` is the correct name.

martindurant:
@tsktsktsk123 do you still have interest in pushing this in, given all versions have since been updated, and the conversation above

tsktsktsk123:
> 
> 
> @tsktsktsk123 do you still have interest in pushing this in, given all versions have since been updated, and the conversation above

Apologies for the delayed answer, been on a holiday.
Regarding the example, I'm still interested in pushing this but I wonder about the relevance. It's been a while since I've been working with RPY but I recall reading that RPy is not able to work multiple instances in parallel.  I'll have to check this, or if someone with the knowledge is able to chime in that'll be great.

Let me know what alterations I need to do and I'll be to process them.

jakirkham:
cc-ing @quasiben (in case this is of interest ;)

quasiben:
@tsktsktsk123 I would recommend installing r-base from `defaults` . This will increase the image size and will be slow to load the first couple times but the loading should decrease over time.  rpy2 is super cool and I think this would be a great example to show off dask interoperating with other languages

jlevy44:
I'm looking to implement in a package I'm working on. Would each worker be able to inherit the R libraries that have been loaded? I worry that it may take too long to load each library again to process each row.

TomAugspurger:
@jlevy44 is that related to this example notebook?

jlevy44:
No, it is not. I suppose I’ll open an issue for this point

jacobtomlinson:
Given the age of this PR and lack of activity for some time I'm going to close it out. Happy to revisit this if folks have the time/inclination.

PR #27, opened 2018-08-01
Additional ml examples

TomAugspurger:
This adds dedicated notebooks for CPU bound (distributed scikit-learn) and RAM bound. I've opted to remove the current top-level `machine-learning.ipynb`. I think it can still serve as a nice high-level overview of some of the main features of Dask-ML. I've attempted to link to the `machine-learning` directory from that notebook. It works locally, so hopefully it will on binder as well.

I've also added a parallel prediction notebook.

### DIFF
--- a/machine-learning.ipynb
+++ b/machine-learning.ipynb
@@ -4,16 +4,10 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "# Dask for Machine Learning"
-   ]
-  },
-  {
-   "cell_type": "markdown",
-   "metadata": {},
-   "source": [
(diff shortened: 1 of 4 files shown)


TomAugspurger:
@mrocklin any objection to me merging this and https://github.com/dask/dask-ml/pull/329? It'll probably take me a few iterations until everything is looking nice in both places.

mrocklin:
Sure, as you like.  I'd like to add the constraint that people clicking on
the "Try Now" link on dask.pydata.org always get redirected to a nice place.

TomAugspurger:
Agreed. I'd like to explore things a bit, but in theory I like the idea of
"top-level" overview notebooks, and then sub-directories for more detailed
examples. The link from dask.pydata.org will continue to point to the
top-level overview notebooks.

PR #33, opened 2018-08-29
Add TPOT example

TomAugspurger:


TomAugspurger:
Rendered version: http://nbviewer.jupyter.org/github/dask/dask-examples/blob/eaf26a7d5de373a119836494667dba6969ee6ce2/machine-learning/tpot.ipynb

mrocklin:
That's a pretty slick video.  I enjoyed hearing about what worked well and
what didn't work well.

TomAugspurger:
This is passing now, with dask-ml being pulled from conda-forge. Does that work for you @mrocklin? Otherwise we can wait till defaults is updated.

mrocklin:
Adding conda-forge for a while seems fine to me.

One comment on the contents of the notebook itself.  They currently assume that the reader is familiar with TPOT and Auto-ML.  I suspect that this will rarely be true.  I think that this could be addressed in two ways:

1.  Add a brief summary of what TPOT does in the header cell
2.  Rename and retitle the notebook to something like auto-ml-with-tpot.ipynb / Automated Machine Learning with TPOT

TomAugspurger:
Updated, merging later today.

PR #34, opened 2018-08-30
Embarrassingly parallel workload example

### LINKED ISSUE #30: Add example for embarrassingly parallel parametrized computation
Would this be of interest for people here to have an example with Dask showing how to do embarrassingly parallel computation on some simulation code?

The idea is that you have a code that does some heavy computation, it takes some parameters and return one or more values as a result. You want to evaluate the program on a grid, with a predefined list of inputs, or maybe with some randomly picked input (Montecarlo simulation). The code run in a few seconds or minutes with each input, and you've got to run it with several thousand or more different inputs. It is something that is often seen in HPC with some physical simulations, and this is not really efficient doing it only with job arrays or alike in PBS or some other schedulers.

Basic steps would be:
1. Generate or read a list of parameters
2. Use client.map or delayed to iterate over these parameters and run the computation
3. Gather all results, and save them on a tabular file.
4. Eventually do some analysis on the result, mean or other aggregations.

I'm currently working on something like that for showing it to our scientists.

Any thought on that?

guillaumeeb:
Closes #30 

First try for me on english example, be merciful on grammar and other language errors. But every feedback is welcome.

### DIFF
--- a/embarrassingly-parallel-with-delayed-or-futures.ipynb
+++ b/embarrassingly-parallel-with-delayed-or-futures.ipynb
@@ -1,0 +1,447 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "<img src=\"http://dask.readthedocs.io/en/latest/_images/dask_horizontal.svg\" \n",
+    "     width=\"30%\" \n",
+    "     align=right\n",
+    "     alt=\"Dask logo\">\n",
+    "\n",
+    "Embarrassingly parallel Workloads\n",
+    "--------------------------------------------------\n",
+    "\n",
+    "This notebook shows using [dask.delayed](http://dask.pydata.org/en/latest/delayed.html) or Futures interface to parallelize generic Python code. \n",
+    "\n",
+    "This example focuses on using Dask for building large embarrassingly parallel computation as often seen in scientific communities and on High Performance Computing facilities, for example with Monte Carlo methods. This kind of simulations suppose the following:\n",
+    " - we have a function that run a heavy computation given some parameters,\n",
+    " - we need to compute this function on a lot of different input parameters, each function call being independant\n",
+    " - we want to gather all the results in one place for further analysis."
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Start Dask Client for Dashboard\n",
+    "\n",
+    "Starting the Dask Client will provide a dashboard which \n",
+    "is useful to gain insight on the computation.  We will also need it fot the\n",
+    "Future API part of this example. Moreover, as this kind of computation\n",
+    "is often launched on super computer or in the Cloud, you will probably end\n",
+    "up having to start a cluster and connect a client. See \n",
+    "[dask-jobqueue](https://github.com/dask/dask-jobqueue),\n",
+    "[dask-kubernetes](https://github.com/dask/dask-kubernetes) or \n",
+    "[dask-yarn](https://github.com/dask/dask-yarn) for easy ways to achieve this.\n",
+    "\n",
+    "The link to the dashboard will become visible when you create the client below.  We recommend having it open on one side of your screen while using your notebook on the other side.  This can take some effort to arrange your windows, but seeing them both at the same is very useful when learning."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "from dask.distributed import Client, progress\n",
+    "client = Client(threads_per_worker=4, n_workers=1)\n",
+    "client"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Define your computation calling function\n",
+    "\n",
+    "This function do a simple operation: adding all numbers of a list/array together, but it sleeps for a random amount of time to simulate real work. In real use cases, this could call another python module, or even run an executable using subprocess module."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "import time\n",
+    "import random\n",
+    "\n",
+    "def costly_simulation(list_param):\n",
+    "    time.sleep(random.random())\n",
+    "    return sum(list_param)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "We can try it"
+   ]
+  },
+  {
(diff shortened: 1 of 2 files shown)


mrocklin:
Thanks for submitting this.  I'm taking a look now.

mrocklin:
I've gone through things and pushed up some changes.  Are you ok with these @guillaumeeb ?

Also, I think that we should find a shorter name for the notebook.

guillaumeeb:
I'm perfectly fine with your changes which makes this example much better looking!

Agreed on the shorter name too, not really good at names :).

mrocklin:
embarrassingly-parallel?  monte-carlo?  simple-parallelism?
parameter-sweep?

guillaumeeb:
I would prefer one of the first two propositions.

mrocklin:
Happy either way.  I guess the first seems more natural to me, though I may
not be the target audience.  Up to you.

guillaumeeb:
Away from keyboard for now. I prefer first  one too.

embarrassingly-parallel-with-delayed-or-futures.ipynb
```diff
@@ -0,0 +1,512 @@
...
+   "metadata": {},
+   "source": [
+    "## Start Dask Client for Dashboard\n",
+    "\n",
+    "Starting the Dask Client will provide a dashboard which \n",
+    "is useful to gain insight on the computation.  We will also need it for the\n",
+    "Futures API part of this example. Moreover, as this kind of computation\n",
+    "is often launched on super computer or in the Cloud, you will probably end\n",
+    "up having to start a cluster and connect a client to scale.  See \n",
+    "[dask-jobqueue](https://github.com/dask/dask-jobqueue),\n",
```
TomAugspurger: Perhaps note that the specific library you want to use depends on your specific compute infrastructure?

embarrassingly-parallel-with-delayed-or-futures.ipynb
```diff
@@ -0,0 +1,512 @@
...
+    "[dask-kubernetes](https://github.com/dask/dask-kubernetes) or \n",
+    "[dask-yarn](https://github.com/dask/dask-yarn) for easy ways to achieve this.\n",
+    "\n",
+    "The link to the dashboard will become visible when you create the client below.  We recommend having it open on one side of your screen while using your notebook on the other side.  This can take some effort to arrange your windows, but seeing them both at the same is very useful when learning."
```
TomAugspurger: typo: "at the same **time** is very useful"

embarrassingly-parallel-with-delayed-or-futures.ipynb
```diff
@@ -0,0 +1,512 @@
...
+    "Note that this is not very clever as we can easily parallelize code. \n",
+    "\n",
+    "There are many ways to parallelize this function in Python with libraries like `multiprocessing`, `concurrent.futures`, `joblib` or others.  These are good first steps.  Dask is a good second step, especially when you want to scale across many machines.\n"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Use Dask Delayed to make our function lazy\n",
```
TomAugspurger: At some point in each section, it would be helpful to state when this approach should be preferred compared to the rest. For example,

> Using `dask.delayed` is a relatively straightforward way to parallelize an existing code base, even if the computation isn't embarrassingly parallel like this one. It's less useful if you need to be calling delayed functions from other delayed functions (the Futures API is better for that kind of workload). See [delayed best practices](http://dask.pydata.org/en/latest/delayed-best-practices.html) for more on using `dask.delayed` effectively.

Basically, we're presenting 3 ways of doing roughly the same thing, so to the extent possible we should give general guidelines on which one is best for common workloads.

embarrassingly-parallel-with-delayed-or-futures.ipynb
```diff
@@ -0,0 +1,512 @@
...
+    "results[:5]"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Using Futures API\n",
+    "\n",
+    "Same example can be done using the Future API by using the client object itself.  For our application of applying a function across many inputs both Dask delayed and Dask Futures are equally useful.  The futures API is a little bit different because it starts work immediately rather than being completely lazy.\n",
```
TomAugspurger: "Same" -> "The same"

TomAugspurger: "Future API" -> Dask's `futures` API?

embarrassingly-parallel-with-delayed-or-futures.ipynb
TomAugspurger: I see now that the Bag section does this, so perhaps my comment can be ignored.

guillaumeeb:
Thanks @TomAugspurger for the review, comments taken into account.

embarrassingly-parallel.ipynb
```diff
@@ -160,10 +161,12 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "## Use Dask Delayed to make our function lazy\n",
+    "## Use [Dask Delayed](http://dask.pydata.org/en/latest/delayed.html) to make our function lazy\n",
     "\n",
     "We can call `dask.delayed` on our funtion to make it lazy.  Rather than compute its result immediately, it records what we want to compute as a task into a graph that we'll run later on parallel hardware.\n",
     "\n",
+    "Using `dask.delayed` is a relatively straightforward way to parallelize an existing code base, even if the computation isn't embarrassingly parallel like this one. It's less useful if you need to be calling delayed functions from other delayed functions (the Futures API is better for that kind of workload). See [delayed best practices](http://dask.pydata.org/en/latest/delayed-best-practices.html) for more on using dask.delayed effectively.\n",
```
mrocklin: > It's less useful if you need to be calling delayed functions from other delayed functions (the Futures API is better for that kind of workload). See [delayed best practices](http://dask.pydata.org/en/latest/delayed-best-practices.html) for more on using dask.delayed effectively

I recommend that we skip this statement for this example.  It's a good point, but I suspect that more beginning readers are likely to be confused by this than gain benefit from it.  They may not understand what it means to call a delayed function from within a delayed function.  Also, this is just as doable/problematic with dask.delayed as calling futures from within futures code.  Both are doable, though neither is particularly recommended.

mrocklin:
I had a small comment but otherwise these changes look good to me.

embarrassingly-parallel.ipynb
guillaumeeb: I'm fine with this, done.

mrocklin:
Merging in.  Thanks @guillaumeeb !

PR #37, opened 2018-09-04
fix to embarassinly parallel notebook

guillaumeeb:
Was trying mybinder, and just run into an issue with this fresh example: we did not run the computation on all inputs despite the documentation saying we were going to do that.
I also noticed that calling `client.map` on the same parameters than `client.submit` before did not trigger any computation.


mrocklin:
Thanks @guillaumeeb

PR #39, opened 2018-09-10
Point to new example layout

ian-r-rose:


mrocklin:
Thanks @ian-r-rose .  Merging.

PR #41, opened 2018-09-14
Add Sphinx documentation with nbsphinx

mrocklin:
This adds basic sphinx quickstart material and then uses nbsphinx to construct
rst pages.

This also slightly modifies many of the notebooks to make them slightly more
uniform in the documentation.

PR #42, opened 2018-09-15
doctr

TomAugspurger:
Waiting on #41


mrocklin:
Rebased and force pushed

mrocklin:
Looks like the build worked well but some final deploy steps failed?

TomAugspurger:
I had the wrong directory for the built docs.

I didn't change anything that would have affected the runtime of the notebooks though. I suppose it's possible the extra verbose output kept travis from timing out the build.

TomAugspurger:
Passing now.

We may want to reduce the verbosity a bit and see if things still pass, but I assume we'll have followup PRs to tweak some things after this is live.

TomAugspurger:
Build at https://travis-ci.org/dask/dask-examples/builds/429316672

Once that finishes, it should automatically create a `gh_pages` branch.

Then we can add `examples.dask.org` as the domain through the github settings.

TomAugspurger:
Here's the built example.

PR #44, opened 2018-09-17
Static site fixups

### LINKED ISSUE #43: Static Site Feedback
http://dask.org/dask-examples/

- Some of the code examples are a bit wide, causing a horizontal scrollbar
- The Client repr could be improved with some external CSS
screen shot 2018-09-16 at 3 15 11 pm
- dask.array output repr is a bit too wide 
screen shot 2018-09-16 at 3 16 27 pm
- pandas HTML table is a bit ugly, some output too long (dataframe.html) 
screen shot 2018-09-16 at 3 17 11 pm
- Section headers feel a bit squished into the previous content, too far from the section content 
screen shot 2018-09-16 at 3 17 57 pm
- Our `code` bocks have the yellow background we've removed in a few places IIRC 
screen shot 2018-09-16 at 3 19 29 pm
- http://dask.org/dask-examples/dataframes/01-data-access.html isn't executing, I think because the video which needs to be check-in as executed. Need to set `nbsphinx_execute = 'always'` in our conf.
- http://dask.org/dask-examples/machine-learning/parallel-prediction.html update links to be absolute links to the API pages, not `:class:` refs
- Empty cell at the bottom of http://dask.org/dask-examples/machine-learning/xgboost.html
- Long output in http://dask.org/dask-examples/embarrassingly-parallel.html#Using-the-Futures-API
- Empty cell at the bottom of http://dask.org/dask-examples/embarrassingly-parallel.html#Doing-some-analysis-on-the-results
- Section headers off in http://dask.org/dask-examples/embarrassingly-parallel.html. The top one should be an `#` and the rest `##`.

TomAugspurger:
Closes #43 

This gets mosts of the things from #43. I haven't touched the Client repr yet. Don't plan to in the near-term.

This should also build with the updated CSS from dask-sphinx-theme.

### DIFF
--- a/machine-learning/scale-scikit-learn.ipynb
+++ b/machine-learning/scale-scikit-learn.ipynb
@@ -15,7 +15,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": null,
+   "execution_count": 1,
    "metadata": {
     "keep_output": true
    },
@@ -29,7 +29,7 @@
        "<IPython.core.display.HTML object>"
       ]
      },
-     "execution_count": null,
+     "execution_count": 1,
      "metadata": {},
      "output_type": "execute_result"
     }
@@ -60,7 +60,7 @@
    "source": [
     "## Distributed Training\n",
     "\n",
-    "<img src=\"images/scikit-learn-logo-notext.png\"/> <img src=\"images/joblib_logo.svg\" width=\"20%\"/> \n",
+    "<img src=\"../images/scikit-learn-logo-notext.png\"/> <img src=\"../images/joblib_logo.svg\" width=\"20%\"/> \n",
     "\n",
     "Scikit-learn uses [joblib](http://joblib.readthedocs.io/) for single-machine parallelism. This lets you train most estimators (anything that accepts an `n_jobs` parameter) using all the cores of your laptop or workstation.\n",
     "\n",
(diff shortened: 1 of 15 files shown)


machine-learning/scale-scikit-learn.ipynb
```diff
@@ -59,7 +60,7 @@
    "source": [
     "## Distributed Training\n",
     "\n",
-    "<img src=\"images/scikit-learn-logo-notext.png\"/> <img src=\"images/joblib_logo.svg\" width=\"20%\"/> \n",
+    "<img src=\"../images/scikit-learn-logo-notext.png\"/> <img src=\"../images/joblib_logo.svg\" width=\"20%\"/> \n",
```
mrocklin: This is tricky.  Jupyter/Jupyterlab would prefer that we keep this as it was.  We'll run into this in other notebooks as well.

mrocklin:
Looks good to me

machine-learning/scale-scikit-learn.ipynb
TomAugspurger: Just to be clear, this is in the `machine-learning/` sub-directory, so the path to the images directory is up one. Do you still think jupyter wants it as is? Because it's rendering correctly for me with the `../`

mrocklin: Well in that case lets leave it as is :) 

+1 from me to merge (though it think that short term we should just merge without coordination).

PR #45, opened 2018-09-18
Added text vectorization example

TomAugspurger:
Moved from dask-ml.


mrocklin:
I changed the title a bit to match the style of the others in this section (currently all active verbs).  We should probably think about a basic set of style guidelines for consistency among examples at some point.

More importantly, when I tried to run the notebook it failed to fetch the reuters dataset.  Maybe it's just that that site is down temporarily.  I'm not sure.

TomAugspurger:
Seems like it is down at the moment.

The recommend way going forward (0.20+) is `sklearn.datasets.fetch_*` so
`fetch_20newsgroups` for this I think.

I'm fine with waiting until 0.20 is out.

mrocklin:
0.20 is out

mrocklin:
It might be helpful to complete a workflow here by including training with some simple estimator.

martindurant:
This should probably be finished or closed.

TomAugspurger:
Thanks for looking through. Updated to

* Remove the custom data fetching / loading, in favor of `sklearn.datasets.fetch_newsgroups20`. This has the downside of not really being an out-of-core example anymore, since the full dataset is loaded into memory by scikit-learn. But I think for the clarity of the example this is an OK tradeoff. I've noted that on real problems you'd replace `dd.from_pandas` with functions to load data from disk using `dask.bag` or `dask.delayed`.
* More links back to our docs
* Removed a bit on unknown chunk sizes. I kept one note about it because it shows up in the array repr, and I want to explain the need for `assume_equal_chunks` in `Incremental`.
* Strip the output 

Will open up an issue about the metadata (the nbytes computation was also incorrect for sparse arrays).

TomAugspurger:
Merging this now, thanks for the reviews.

mrocklin:
Thanks Tom!

PR #47, opened 2018-09-27
Update scikit-learn to 0.20

mrocklin:
This allows us to remove some recommendations around joblib and rely on automated behaviors

cc @TomAugspurger @ogrisel


machine-learning.ipynb
```diff
@@ -114,7 +119,7 @@
    "source": [
     "from sklearn.externals import joblib\n",
     "\n",
-    "with joblib.parallel_backend('dask', scatter=[X, y]):\n",
```
TomAugspurger: @ogrisel what's the current state of auto-scattering? IIRC, there was a performance issue? I may be misremembering.

ogrisel:
Sorry for the late feedback: auto-scattering is enabled by default and should work almost as fast as explicit scattering. +1 for removing explicit scattering from the examples.

Issue #20, opened 2018-07-14
Add example for dealing with satellite imagery


mrocklin:
[This notebook](https://gist.github.com/mrocklin/3df315e93d4bdeccf76db93caca2a9bd) demonstrates using XArray and Dask with a large satellite image.  It downloads the image from S3 with rasterio and then loads it in chunks.  

It doesn't do a whole lot with it afterwards though, which presumably we should change.  This, along with ensuring that the computation we choose also fits nicely in RAM (might have to play around with chunk sizes a bit) is presumably the hard part of this exercise.

Additionally, there is an open question of whether or not we should include rasterio in the docker image (I suspect that it brings in GDAL, which would greatly increase the image size).  Instead we might include a `!pip install xarray rasterio` line or something similar at the top of the notebook.

guillaumeeb:
I'm interested in this example, maybe I can work on it if I have find some time, this is typically something which can be of interest at CNES.

I can propose a basic example on satellite imagery which is NDVI computation, it just need an image with NIR band, which can be found in Sentinel 2 images.

See https://developers.google.com/earth-engine/tutorial_api_06 for an example.

We could compute that and display it in a plot.

mrocklin:
That would be quite welcome

guillaumeeb:
I think this issue is closed by #60!

mrocklin:
Indeed.  Thanks @rmsare !

PR #60, opened 2019-02-18
Add example of calculation with satellite imagery

rmsare:
Following on #20, this adds a simple example of downloading, transforming, and calculating a derivative index from a Landsat image using dask and xarray. I adapted it from my usual workflow, so I'd appreciate any feedback and would be happy to improve it.

In particular, I feel there is too much application-specific helper code. On the other hand, it is nice to show that pre-processing can be made more efficient with dask.


rmsare:
Oops! Poking around I noticed that this example is similar to what was started in [images-ndvi.ipynb](https://github.com/guillaumeeb/dask-examples/blob/images_ndvi/images-ndvi.ipynb). 

I didn't mean to conflict with any work in progress but happy to contribute if this issue is still up for grabs.

martindurant:
This may not be the place to say so, but the first few cells up to open_rasterio could nicely be encoded in a catalogue for Intake: intake-xarray supports rasterio, and download-on-first-use.

mrocklin:
> Oops! Poking around I noticed that this example is similar to what was started in [images-ndvi.ipynb](https://github.com/guillaumeeb/dask-examples/blob/images_ndvi/images-ndvi.ipynb).

> I didn't mean to conflict with any work in progress but happy to contribute if this issue is still up for grabs.

cc @guillaumeeb who may know more about current status there or want to collaborate.

mrocklin:
> This may not be the place to say so, but the first few cells up to open_rasterio could nicely be encoded in a catalogue for Intake: intake-xarray supports rasterio, and download-on-first-use.

@martindurant are you suggesting that contributors to this repository should start to use intake as a common practice in their notebooks?

martindurant:
I can understand that you probably don't want any dask examples to depend on Intake (they should be demonstrating dask, not intake), but if this were anywhere else, then yes, a notebook like this one I think could be cleaner with intake and is just the kind of workflow it was made for.
Eventually, if Intake is established, then I could imagine it making a lot of sense to have the datasets used in dask-examples be formulated via an intake catalogue.

rmsare:
@martindurant Thanks for pointing that out. I was totally unfamiliar with Intake but the Landsat catalog example is really nice! I'm going to start incorporating it into my personal work.

martindurant:
(cc @jsignell , who worked on landsat)

mrocklin:
Thanks generally for the contribution @rmsare !

In general I think that it would be great to have more Xarray style examples in examples.dask.org, at least for as long as an examples.xarray.org does not yet exist.  In my opinion the notebook you've authored here is nicely self-contained, easy to run through, and well explained.  I enjoyed reading through it.  

One *technical* challenge I can imagine us running into is that the software environment that we provide to users to run these notebooks is unlikely to have `rasterio` in it, largely due to the large GDAL dependency.  My guess is that adding `rasterio`/`GDAL` to the environment would increase it by a few hundred megabytes, which in turn slows down all users (we've tried to keep things somewhat lean).  If my assumption that adding rasterio to the environment would cause non-trivial bloat is correct, then we have some options:

1.  Install rasterio in the first cell, with something like

        !conda install rasterio

    We do this for other small dependencies, usually using pip (because it's faster).  I'm not sure what the experience would be like for rasterio.  It would be worth trying this out on a binder notebook if you have the time.

2.  Xarray could set up its own examples page and we could move this work there.  cc @jhamman 
3.  We could expand examples.dask.org to include non-runnable examples that were only viewable on the static site, but not tested or runnable

rmsare:
@mrocklin Thanks - didn't realize xarray is in the default environment. I'm testing things out in a binder NB now.

mrocklin:
Xarray on its own is small and easy to keep around.  Rasterio/GDAL not so much :)

rmsare:
Yep, GDAL is a monster. Running the full notebook on binder seems to work, with a reasonable install time for rasterio and everything else working fine. 

Not surprisingly executing the final call to `ndvi2d.plot` does take a while (> 30 s) but no timeout. Occasionally the kernel dies but I can't reproduce that consistently.

https://mybinder.org/v2/gh/rmsare/dask-examples/satellite-imagery

guillaumeeb:
I'm really pleased that someone took over this. Sorry for not being able to move forward myself. I hope I'll find some time to test and review!

applications/satellite-imagery-geotiff.ipynb
```diff
@@ -0,0 +1,479 @@
...
+    "## Check image metadata\n",
+    "\n",
+    "Let's see if the image is tiled so we can select a chunk size."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 6,
+   "metadata": {},
+   "outputs": [
```
guillaumeeb: IIRC we decided to clear all outputs from our example in order to keep things small and avoid diffs.

applications/satellite-imagery-geotiff.ipynb
```diff
@@ -0,0 +1,479 @@
...
+       "  * band     (band) int64 1"
+      ]
+     },
+     "execution_count": 11,
+     "metadata": {},
+     "output_type": "execute_result"
+    }
+   ],
+   "source": [
+    "red.sum(dim=['x', 'y']).compute()"
```
guillaumeeb: Do this operation has sense in some cases? Do we want to keep it or give a more meaningfull example?

applications/satellite-imagery-geotiff.ipynb
```diff
@@ -0,0 +1,479 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Rescale bands using Landsat metadata\n",
+    "\n",
+    "The Landsat Level 1 images are delivered in a quantized format. This can to be [converted to top-of-atmosphere reflectance](https://landsat.usgs.gov/using-usgs-landsat-8-product) using the provided metadata.\n",
```
guillaumeeb: Typo : this has to be.

applications/satellite-imagery-geotiff.ipynb
```diff
@@ -0,0 +1,479 @@
...
+      "array([0.059213])\n",
+      "Coordinates:\n",
+      "  * band     (band) int64 1\n"
+     ]
+    }
+   ],
+   "source": [
+    "print(red.max(dim=['x', 'y']).compute())\n",
+    "print(red.min(dim=['x', 'y']).compute())\n",
+    "print(red.mean(dim=['x', 'y']).compute())"
```
guillaumeeb: Could we do all these operations into only one aggregation with several operations?

applications/satellite-imagery-geotiff.ipynb
```diff
@@ -0,0 +1,479 @@
...
...4DGNvesIsQzDYijg5NdZslCCdZNWgt4WYbj4junC51hqhbAwTCZTqKzHlkv0uoDipLrFttsg5Pxw5XeALsMpY2wqPHokw0EhhzfeOUmVVIHaWMzM9FHt+KSde3Elrr+7TWe7XWu7XWv5k99l6t9T3f4NiXvxhVClxKKhd3eJPZYpJVJ9oki65Q4LhosYgx6DKoMLnVDhMGOVRxcxxub3DfzAmmckXKUtHQLkInNJQkVCmeNLa0kVY0kpibhiYo2Q77+wZ4vLFKLHweWJ3HtywSNJN+kb9fOMWB6hBnNjeQQtLnuKRoIqXYW64a06himUQpcpbN3585xkqnzZH1ZXzbZj0MeHxhltONdXzb5rMzp/h8kOfExipD/pYk+lTfAAvtTX7t85/mH2dOoLRiPexwz/wpPrK0Tp/lcHj5HO+Y2Ea/7fKuqSsZ8gvcNbHbmIwtz1P1fL5vaJwj68ts8/J899gulqMOnTShartsxCG2kOws93OwWCEnJNu8vOGkyDwSerOTcStmLmzzYH0RV0hOd5qE2TrpZSrJQNYKkoQYDgjCZtKBqt4wFUGyBhlE+mu8Urpkx+4GoTtT6cqwCIm2TFWHXcm+p7BUxyDMsvZVaBml4Z7isHAz+K8Cp5YpGTd6LTdLJ1T0Jj4JO5zzzkG1QbVxVMtUX9LtzX2u9AugFatJgs7mQPs8I5svRWZlnNTRTg3cEapsEp/4wIv9KbngwuhKf/PbBUyov5RULu5QW1+7CDCdZGq1Zpc5zEbWF892qTrqJZa1JGI4X+TMZp3XbpviyU0j51FPIpoaytKgUGbCNs+0N9jvl9lMEwbdHEuhcQQ8UKzgCMHtA6Nm2IrgXBhwZf8QtVyeiaLxR5luNYjSlMsrNTxpsRC0WOq0DfvdcQmimMXNzZ4R17XVEW4en+SawVGOr6+yp3+ApaDFtbVRjqwvs6vcT8Fx6PdyuNKiGXQoux5BmnDFwDA52+EtNZ+WSnj3ZdeiGk9z6x//PvUkIk6NRlWQphysDrEZm9f7lm17+IuMy7LTLzKZL/F8sMm+Qh/zQROV2QqIbIFX2fzEE4Yl7Uubtigw6dpc2zfIUhwy7uV7/BmARmo0rdzsvfIEW4RFoCErrJI37yEKr+uP0k0m0JPi6Wl3ddnxmReJiFdMQsgqmFiWzDHnPeZlbdCoOzMRmSqwdGnrbKbijmxtVn
```
guillaumeeb: In my case, computation ends, but plot never showed... Upon recomputation, notebook failed. Sorry I did not have time to investigate more....

mrocklin:
Thanks for taking the time to review @guillaumeeb !

applications/satellite-imagery-geotiff.ipynb
rmsare: Thanks for these comments, I noticed this too. Usually for me, re-running the cell results in the plot showing, or rarely the kernel dies. 

I will debug this further today and see if I can get it to execute consistently.

applications/satellite-imagery-geotiff.ipynb
rmsare: Oops, done!

applications/satellite-imagery-geotiff.ipynb
rmsare: Thanks, fixed

applications/satellite-imagery-geotiff.ipynb
rmsare: Fair point, maybe for this case it would be more realistic to compute a windowed mean, resampling the image. I kept this mostly because it has a simple execution plot in the dashboard.

applications/satellite-imagery-geotiff.ipynb
rmsare: Fixed this by using `%matplotlib inline`

applications/satellite-imagery-geotiff.ipynb
mrocklin: I've gone ahead and done this.

mrocklin:
This notebook is live here if people want to play with it in the binder environment.

https://mybinder.org/v2/gh/rmsare/dask-examples/satellite-imagery?urlpath=lab/tree/applications/satellite-imagery-geotiff.ipynb

rmsare:
@guillaumeeb Thanks for your review! and @mrocklin, I'm sorry it took me a while to get back to this. I deleted the first example per the review.

Based on discussion above, happy to contribute this as a runnable example on examples.dask.org or wait and re-submit to a separate xarray page, whichever makes sense.

applications/satellite-imagery-geotiff.ipynb
rmsare: Deleted this example

applications/satellite-imagery-geotiff.ipynb
```diff
@@ -0,0 +1,334 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "red = calculate_reflectance(red, band_number=4)\n",
+    "nir = calculate_reflectance(nir, band_number=5)"
```
jsignell: Slightly confusing to overwrite original meaning of `red` and `nir`. Might be better to use new variable names like `red_toa` or `nir_toa`.

applications/satellite-imagery-geotiff.ipynb
```diff
@@ -0,0 +1,334 @@
...
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "red_max, red_min, red_mean = dask.compute(\n",
+    "    red.max(dim=['x', 'y']), \n",
+    "    red.min(dim=['x', 'y']),\n",
+    "    red.mean(dim=['x', 'y'])\n",
+    ")\n",
+    "print(red_max)\n",
```
jsignell: Makes it easier to read the output if you only print the float. 

```suggestion
    "print(red_max.item())\n",
```

applications/satellite-imagery-geotiff.ipynb
```diff
@@ -0,0 +1,334 @@
...
+    "print(red_min)\n",
```
jsignell:
```suggestion
    "print(red_min.item())\n",
```

applications/satellite-imagery-geotiff.ipynb
```diff
@@ -0,0 +1,334 @@
...
+    "print(red_mean)"
```
jsignell:
```suggestion
    "print(red_mean.item())"
```

applications/satellite-imagery-geotiff.ipynb
```diff
@@ -0,0 +1,334 @@
...
+    "ndvi = (nir - red) / (nir + red)"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "ndvi2d = ndvi.isel(band=0)"
```
jsignell: I think it is clearer to use `squeeze` in this case since the band selection doesn't really mean anything:

```suggestion
    "ndvi2d = ndvi.squeeze()"
```

applications/satellite-imagery-geotiff.ipynb
rmsare: I agree, fixed

applications/satellite-imagery-geotiff.ipynb
rmsare: Oops, should have caught this. Changed.

applications/satellite-imagery-geotiff.ipynb
rmsare: I agree! Thank you for these comments.

mrocklin:
Thanks @jsignell and @guillaumeeb for the review.  OK to merge?

jsignell:
LGTM

mrocklin:
Great.  Thanks all!

Thanks for the contribution @rmsare !  It should be fun to play with.  This should be live on https://examples.dask.org shortly.

PR #61, opened 2019-02-27
Add quickstart guide to dask-image

GenevieveBuckley:
Currently there is no quickstart guide to using dask-image. @jakirkham suggests that this repository is the best place to add such a notebook.

I've added:
1. A jupyter notebook `dask-image-quickstart,ipynb`, and 
1. Some small public domain image files to use in the example notebook.

I've also made two changes to the environment.yml file mybinder uses:
1. Added dask-image
2. Replaced the pip installation of graphviz with the conda-forge installation of python-graphviz (which contains graphviz as a dependency). The dask-image-quickstart notebook visualizes the computational graph in several places, and it's my understanding that I need python-graphviz, not just graphviz to do that. I don't believe this will cause any issues elsewhere.

I've confirmed the notebook runs on mybinder.org. Please let me know if there's anything else I've forgotten to do here.


mrocklin:
Hi @GenevieveBuckley !  Thanks for submitting this.  I think that this would be a great notebook to include.  I would like for us to make a few changes though if possible:

1.  We're trying to keep this repository small (given people's desires to include datasets it can grow quite large otherwise).  So can you
    1.  Strip the notebook of outputs
    2. Remove the images from the repository.  Perhaps we can download them from the web in the first cell of the notebook instead?

    Collectively these should reduce the size of the commit considerably
2.  Use absolute imports for `dask_image`

    ```diff
    -from dask_image import imread
    -x = imread(...)
    +import dask_image
    +x = dask_image.imread(...)
    ```

    I think that this is especially valuable because this notebook will be read by a larger audience, who isn't necessarily focused on dask_image.  When I first saw `ndfilters` in the code I thought that it was from scipy and had to scroll up to the imports to figure out where it was coming from.

mrocklin:
Additionally, do you have any thoughts on other sets of images perhaps?  I wonder if there is some set of a few images that make up a stack more naturally.  Splitting Eileen in four and then stacking those quandrants is sensible, but not something that I imagine looks like what people do in practice.  I wonder if there is some small scientific dataset out there that makes sense as a genuine stack.  No worries if there isn't anythign convenient that fits the bill here, what we have here is fine.  I'm just throwing out ideas.

GenevieveBuckley:
1. i. Done. Notebook output is cleared.
1. ii. Done. Data downloaded and saved to temporary directories instead. (I think it's really important for new users to practise loading data from disk into dask arrays. Temporary directories seem like an ok way to achieve this.)
2. Done. 

### Re: thoughts on dask-image example data.
@jakirkham and I have been talking about ideas for this, although I'm not sure we're any closer to a good solution. I was about to link you to the issue for this discussion in dask-image... and it appears there isn't one. Oops.

I've just opened an issue for this in dask-image, let's shift this discussion there: https://github.com/dask/dask-image/issues/107

dask-image-quickstart.ipynb
```diff
@@ -0,0 +1,832 @@
...
+    "We first install the library scikit-image for easier access to the example image data there."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "!pip install scikit-image"
```
mrocklin: We're getting the error we get if we update dask to the most recent version.  I suspect that scikit-image is updating dask for us?

Maybe adding `--upgrade-strategy only-if-needed` or `--no-deps` would help here.

mrocklin:
> I've just opened an issue for this in dask-image, let's shift this discussion there:

Subscribed.  I look forward to seeing what happens there.

dask-image-quickstart.ipynb
```diff
@@ -0,0 +1,832 @@
...
+   "source": [
+    "<a id='embarrassingly_parallel'></a>\n",
+    "### Embarrassingly parallel problems"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "The xyntax is identical to apply a function to multiple images or dask chunks. This is an example of an embarrassingly parallel problem, and we see that dask automatically creates a computation graph for each chunk.\n"
```
jakirkham: nit: `xyntax` -> `syntax`

dask-image-quickstart.ipynb
jakirkham: Are we comfortable just adding scikit-image as a requirement? That would likely improve the experience both for maintainers and users of this example.

mrocklin: It looks like the conda-forge [scikit-image package is around 23MB](https://anaconda.org/conda-forge/scikit-image/files)

This is larger than I'm super happy with if there is just one notebook.  I think that for as long as there are only one or two such notebooks I'd prefer to have users pip install in the notebook.  If we invest more heavily here and these notebooks gets a lot of visibility then I think that we would add scikit-image to the docker image.  This is on the border though, if you all feel strongly I'm happy to be overruled. 

This is what we do for other packages like rasterio and tpot.

jakirkham:
Thanks for getting this started @GenevieveBuckley. Also thanks @mrocklin for shepherding this along.

It looks like there are some minor points being discussed above, but these don't appear that difficult to solve.

The only major point to discuss is about sample data for the example. There is actually a fair bit to unpack here. @GenevieveBuckley went into these a bit in issue ( https://github.com/dask/dask-image/issues/107 ). I think some of the challenges with finding suitable data are as follows.

1. License - what data can we use that others can start with easily?
2. Accessibility - can we just download it or must we login? are their constraints on how it may be used?
3. Format - is the data in a reasonably accessible format?
4. Size - does the data work reasonably well on memory poor machines or [Binder]( https://github.com/jupyterhub/binder/blob/master/doc/faq.rst#how-much-memory-am-i-given-when-using-binder )?
5. Generalizability - does the data make sense for users in a broad array of disciplines?
6. Scalability - does the example make sense for multithreaded and distributed use cases?
7. Visualization - is it possible to look at the data and results with some stock visualization tools?

Given these constraints, I think what @GenevieveBuckley has done here is pretty reasonable. Something not too different from this workflow could make sense for users trying to leverage multiple cores in their analysis or trying to perform analysis on a large volume. We may actually want to store the sliced data in this repo to break the setup code out of the example, but that is more or less a separate point.

We can certainly think of other datasets, but they are likely not to meet one of these requirements. For example, we might find a larger dataset, but it is likely to be domain specific (not generalizable), perform poorly on Binder, and possibly have some accessibility and formatting concerns.

If the complaint is the example is somewhat contrived, I can certainly agree with that. However this criticism could be applied to many examples as it is rarely realistic to dive into the complexities one may encounter with real world data when introducing concepts. This is only my opinion though.

mrocklin:
Sounds reasonable to me.  I'm happy to merge what we have here and hope for a better dataset in the future.  

FWIW between the choice of the Astronaut photo and a domain specific image stack (such as from microscopy) I'd personally prefer the domain specific image stack.  As long as the analysis isn't hyper-specific I think that having a domain-specific dataset is fine.

dask-image-quickstart.ipynb
jakirkham: Makes sense. We could vendor the data as noted elsewhere. This may make sense for other reasons too.

It's worth noting that `scikit-image` may already be pulled in as an indirect dependency of `dask-image` currently. Long term we plan to make `scikit-image` an explicit dependency of `dask-image` as there are algorithms we plan to use from it.

jakirkham:
Thanks Matt. That sounds good.

Down the road we may want to add domain specific analyses as well. Ultimately we may need to make some decisions about whether these live here, in dask-image's docs, or in other downstream libraries docs. Happy to discuss in the future.

GenevieveBuckley:
Ok, I think this is ready to merge.

I think I've addressed all the points raised here (with the exception of the longer term discussion about [better dask-image examples](https://github.com/dask/dask-image/issues/107)), and the tests pass. 

@jakirkham we should put something in the dask-image docs (and/or readme) to point people to these examples.

dask-image-quickstart.ipynb
```diff
@@ -0,0 +1,832 @@
...
+   "source": [
+    "<a id='segmenting'></a>\n",
+    "### Segmenting"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "AFter the image preprocessing, we segment regions of interest from the data. We'll use a simple arbitrary threshold as the cutoff, at 75% of the maximum intensity of the smoothed image."
```
jakirkham:
```suggestion
    "After the image preprocessing, we segment regions of interest from the data. We'll use a simple arbitrary threshold as the cutoff, at 75% of the maximum intensity of the smoothed image."
```

jakirkham:
Thanks for working on this, @GenevieveBuckley. Small typo above. Otherwise LGTM

Agree we should link to this example from the docs. Would you be willing to raise an issue or PR at dask-image suggesting a location?

GenevieveBuckley:
> Thanks for working on this, @GenevieveBuckley. Small typo above. Otherwise LGTM

Thanks for catching these typos for me, it's much appreciated.

> Agree we should link to this example from the docs. Would you be willing to raise an issue or PR at dask-image suggesting a location?

Yes I'm willing, but it's at the bottom of my to do list right now. I'll add the issue so it doesn't fall off the radar entirely, and you can follow up later if it doesn't get done.

jakirkham:
Planning on merging tomorrow if no comments.

mrocklin:
I've moved this to `applications/`  and then added it to the index so that it comes into the table of contents.  +1 to merge from me

mrocklin:
This is in.  Thanks for sticking with this @GenevieveBuckley !  I'm quite happy to see this in.

This should be live in a few minutes.

Issue #62, opened 2019-02-27
Maintenence: update mybinder environment.yml dependency versions


GenevieveBuckley:
I noticed that the mybinder environment.yml file pins dask to version 0.20, but the latest dask release is now up to 1.1.2. It's probably time to update or unpin some of these dependencies. Should we do that?

Currently: 
https://github.com/dask/dask-examples/blob/master/binder/environment.yml

```
channels:
  - conda-forge
dependencies:
  - python=3
  - bokeh=0.13
  - dask=0.20
  - dask-ml=0.10.0
  - distributed=1.24
  - jupyterlab=0.35.1
  - nodejs=8.9
  - numpy
  - pandas
  - pyarrow==0.10.0
  - scikit-learn=0.20
  - matplotlib
  - nbserverproxy
  - nomkl
  - h5py
  - xarray
  - bottleneck
  - py-xgboost
  - pip:
    - graphviz
    - dask_xgboost
    - seaborn
    - mimesis
```

mrocklin:
Yes indeed.  That would be good to do.  Would you be willing to submit the
PR?  I'll be curious (and hopeful) to see if the tests pass or if there
were any conflicting API changes.

GenevieveBuckley:
Sure, I can do that. 

Do you prefer to have things unpinned if possible, or just increment to a newer release? We'll probably run into this problem again (outdated dependencies) before too long if we keep things pinned, so I'd prefer not to. But I'd understand if you feel pull requests aren't frequent enough on this repo to trigger tests and alert us of any future issues.

mrocklin:
I'd generally prefer that we keep things pinned.  Otherwise bugs from
mismatched versions might creep in without us knowing about it and things
will go down and cause confusion.  Instead I'd rather that we be explicit
when we change versions and address any bugs that come up when we make such
a change.

GenevieveBuckley:
To update: I think it's better to put this issue on hold until the next dask-ml release is available. 

I think this because upgrading dask versions has unearthed a few bugs, and the new release should contain some of the fixes we need.

### Error 1
`AttributeError: module 'dask' has no attribute 'sharedict'`, seen in the dask-examples/machine-learning/incremental.ipynb notebook. This is addressed by the merged PR https://github.com/dask/dask-ml/pull/455,

### Error 2
[...]

PR #63, opened 2019-03-01
Update dependency versions in binder environment

GenevieveBuckley:
The dependency versions in the binder environment file have fallen reasonably out of date. This pull request brings it up to date.

Version upgrades for pinned dependencies:
```
bokeh=0.13        --> bokeh=1.0.4
dask=0.20         --> dask=1.1.4
dask-ml=0.10.0    --> dask-ml=0.12.0
distributed=1.24  -->  distributed=1.26
```

Closes https://github.com/dask/dask-examples/issues/62

### DIFF
--- a/binder/environment.yml
+++ b/binder/environment.yml
@@ -24,3 +24,4 @@ dependencies:
     - dask_xgboost
     - seaborn
     - mimesis
+    - git+https://github.com/dask/dask@2732793e686dd8e81c967ec54e9fea82b5308226
--- a/binder/environment.yml
+++ b/binder/environment.yml
@@ -3,7 +3,7 @@ channels:
 dependencies:
   - python=3
   - bokeh=1.0.4
-  - dask=1.1.3
+  - dask=1.1.4
   - dask-ml=0.12.0
   - distributed=1.26
   - jupyterlab=0.35.1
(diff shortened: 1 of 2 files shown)


mrocklin:
It looks like there was another small issue with groupby aggregations.  I've resolved this in master and things now pass here.  I'll probably push out another micro-release at the end of the week and then we can merge this in.

jakirkham:
Thanks for helping with this, Genevieve! :)

jakirkham:
So Dask 1.1.4 has landed. Should we update to that? Anything else needed here?

GenevieveBuckley:
> So Dask 1.1.4 has landed. Should we update to that? Anything else needed here?

Dask version updated, thanks! Tests pass.

Two final questions:

1. Should we use conda install wherever possible? As far as I can tell, these packages are all available on conda-forge (at least they are now, not sure about when the environment file was put together):
* `dask_xgboost`
* `seaborn`
* `graphviz`

2. And on that last point, I think we should probably replace this with a conda install of `python-graphviz` from conda-forge, to match with the recommendation [given in the dask docs](http://docs.dask.org/en/latest/graphviz.html). Do other people agree?

mrocklin:
Conda's default `graphviz` package is sometimes broken, which I think is probably why we were using pip and the system library here.  I'll give it a shot and see what happens.

Also, a binder link for this PR is here: https://mybinder.org/v2/gh/GenevieveBuckley/dask-examples/update-binder-env?urlpath=lab

mrocklin:
Looks good.  Merging this in.  Thanks for sticking with this @GenevieveBuckley !

PR #67, opened 2019-04-14
Bump versions

ian-r-rose:
This bumps the JupyterLab and dask-labextension versions. I'm aiming at experimenting with the auto-starting client functionality, but wanted to put that more controversial stuff in a follow-up.

### DIFF
--- a/binder/environment.yml
+++ b/binder/environment.yml
@@ -6,7 +6,6 @@ dependencies:
   - dask=1.1.5
   - dask-image=0.2.0
   - dask-ml=0.12.0
-  - jupyterlab=0.35.4
   - nodejs=8.9
   - notebook<5.7.5
   - tornado=5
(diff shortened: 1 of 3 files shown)


mrocklin:
https://mybinder.org/v2/gh/ian-r-rose/dask-examples/bump-versions?urlpath=lab

Things look nice to me (having the logo replace the 404 is especially nice :) )

I was surprised at how long it took the LocalCluster object to pop up.  That might just be the cost of starting new processes though.  

At some point, if we replace all of the LocalCluster calls in the notebooks with this then we'll want to configure the processes/threads/memory to be a bit smaller probably.  That should probably happen later though.

+1 from me

mrocklin:
Merging tomorrow if there are no follow-up comments.

mrocklin:
Thanks @ian-r-rose !  Looks nice.  Merging.

PR #69, opened 2019-04-24
iamges -> images.

Carreau:


martindurant:
Same typo twice!
Thanks for the correction.

Carreau:
Thanks.

Issue #74, opened 2019-06-05
Add example showing resilience against worker deaths?


willirath:
In preparation for a dask-jobqueue workshop, I'm playing with partially preemptible clusters.  Would it make sense to add an example using a mocked up partially resilient `LocalCluster` to show that dask is able to handle dying ressources?

A draft that works on the current examples binder is here:
https://gist.github.com/willirath/c6e5e26204b33bb4415923329a6b8723

TomAugspurger:
I think that would make for a nice example. Based on the example, my main
comment would be adding a link to
https://distributed.dask.org/en/latest/resilience.html, and noting that a
node getting preempted
is essentially a hardware failure:
https://distributed.dask.org/en/latest/resilience.html#hardware-failures.

TomAugspurger:
I think this was closed by https://github.com/dask/dask-examples/pull/75

PR #75, opened 2019-06-05
Add resilient example

willirath:
[Binder](https://mybinder.org/v2/gh/willirath/dask-examples/add-resilient-example?filepath=resilience.ipynb)

See #74 

ToDo's (so far):

- [x] Link to http://distributed.dask.org/en/latest/resilience.html#hardware-failures and explain preemption as hardware failure (from Dask's PoV)
- [x] ~~Add example dealing with Exceptions in user code? Relevant docs: http://distributed.dask.org/en/latest/resilience.html#user-code-failures~~

### DIFF
--- a/resilience.ipynb
+++ b/resilience.ipynb
@@ -6,7 +6,7 @@
    "source": [
     "# Resilience against hardware failures\n",
     "\n",
-    "Scenario:  We have a cluster that partially consists of preemptible ressources.  That is, we'll have to deal with workers suddenly being shut down during computation.\n",
+    "Scenario:  We have a cluster that partially consists of preemptible ressources.  That is, we'll have to deal with workers suddenly being shut down during computation.  While demonstrated here with a `LocalCluster`, Dask's resilience against preempted ressources is most useful with, e.g., [Dask Kubernetes](https://kubernetes.dask.org/) or [Dask Jobqueue](https://jobqueue.dask.org).\n",
     "\n",
     "Relevant docs: <http://distributed.dask.org/en/latest/resilience.html#hardware-failures>"
    ]
@@ -17,7 +17,7 @@
    "source": [
     "## Increase resilience\n",
     "\n",
-    "Whenever a worker shuts down, the scheduler will increment the suspicousness counter of _all_ tasks that were assigned (not necessarily computing) to the worker in question.  Whenever the suspiciousness of a task exceeds a certain threshold (3 by default), the task will be considered broken.  We want to compute many tasks on only a few workers with workers shutting down randomly.  So we expect the suspiciousness of all tasks to grow rapidly.  Let's effectively disable the threshold by setting it to a large number:"
+    "Whenever a worker shuts down, the scheduler will increment the suspiciousness counter of _all_ tasks that were assigned (not necessarily computing) to the worker in question.  Whenever the suspiciousness of a task exceeds a certain threshold (3 by default), the task will be considered broken.  We want to compute many tasks on only a few workers with workers shutting down randomly.  So we expect the suspiciousness of all tasks to grow rapidly.  Let's increase the threshold:"
    ]
   },
   {
@@ -28,7 +28,7 @@
    "source": [
     "import dask\n",
     "\n",
-    "dask.config.set({'distributed.scheduler.allowed-failures': 1e32});"
+    "dask.config.set({'distributed.scheduler.allowed-failures': 100});"
    ]
   },
   {
@@ -95,13 +95,13 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "N = 5000\n",
+    "N = 400\n",
     "\n",
-    "x = db.from_sequence(range(N), npartitions=N // 10)\n",
+    "x = db.from_sequence(range(N), npartitions=N // 2)\n",
     "\n",
     "mults = x.map(multiply_by_two)\n",
     "\n",
-    "summed = mults.reduction(sum, sum)"
+    "summed = mults.sum()"
    ]
   },
   {
--- a/resilience.ipynb
+++ b/resilience.ipynb
@@ -1,0 +1,415 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Scenario\n",
+    "\n",
+    "We have a cluster that partially consists of preemptible ressources.  That is, we'll have to deal with workers suddenly being shut down during computation."
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## A cluster"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 1,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "from dask.distributed import Client, LocalCluster"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 2,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "cluster = LocalCluster(threads_per_worker=1, n_workers=5, memory_limit=200e6)"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 3,
+   "metadata": {},
+   "outputs": [
+    {
+     "data": {
+      "text/html": [
+       "<table style=\"border: 2px solid white;\">\n",
+       "<tr>\n",
+       "<td style=\"vertical-align: top; border: 0px solid white\">\n",
+       "<h3>Client</h3>\n",
+       "<ul>\n",
+       "  <li><b>Scheduler: </b>tcp://127.0.0.1:36509\n",
+       "  <li><b>Dashboard: </b><a href='/user/dask-dask-examples-n2kj24go/proxy/8787/status' target='_blank'>/user/dask-dask-examples-n2kj24go/proxy/8787/status</a>\n",
+       "</ul>\n",
+       "</td>\n",
+       "<td style=\"vertical-align: top; border: 0px solid white\">\n",
+       "<h3>Cluster</h3>\n",
+       "<ul>\n",
+       "  <li><b>Workers: </b>5</li>\n",
+       "  <li><b>Cores: </b>5</li>\n",
+       "  <li><b>Memory: </b>1000.00 MB</li>\n",
+       "</ul>\n",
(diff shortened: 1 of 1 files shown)


mrocklin:
Some small notes:

1.  Lets merge cells together when we can.  For example the first three that import, make a cluster, make a client.  The user doesn't gain anything by pressing Shift-Enter three times rather than once.
2.  The memory limit is small.  If they just `import pandas` then we'll probably be close to the limit, which will raise lots of annoying error messages.
3.  `cluster.scheduler.allowed_failures = int(1e32)` This is internal API.  Maybe use `dask.config.set({'distributed.scheduler.allowed-failures': 1e32})` before creating the cluster instead?
4.  The function `_get_worker_pids` seems to be a one-liner that is called once.  Lets just call the code directly instead of hiding it in a function.
    ```python
    pids = [w.pid for w in cluster.scheduler.workers.values()]
    pids
    ```
5.  Same with `_get_preemptible_worker_pids`.  Also, lets change the filter into a list comprehension, which seems to be more broadly understood
7. It's taking me a while to understand `maybe_kill_n_perc_of_workers_and_wait`, which means that a new reader probably doesn't have much chance.  There are too many function calls deep for me to trace through what is happening directly.

    I wonder if we might be able to do something with fewer steps of indirection like the following. 

    ```python
    def kill_a_worker():
        worker = random.choice(cluster.scheduler.workers.values())
        os.kill(worker.pid)

    summed = client.compute(summed)

    while not summed.finished():
        kill_a_worker()
        sleep(...)
    ```

    This is subjective (code is always way simpler to the author than the reader) but I *think* that I could put this in front of someone and they would have a better chance of tracking things through in a 10-20s attention span.

willirath:
Thanks @mrocklin for these remarks.  I have addressed them as far as possible.

There's a few things that I couldn't, however, simplify as much as I'd have liked to:

- While somewhat more streamlined now, the logic of finding a worker to kill and leaving some of them running all the time is now this:

```python
_all_current_workers = [w.pid for w in cluster.scheduler.workers.values()]
non_preemptible_workers = _all_current_workers[:2]
```
```python
def kill_a_worker():
    preemptible_workers = [
        w.pid for w in cluster.scheduler.workers.values()
        if w.pid not in non_preemptible_workers]
    if preemptible_workers:  # need this  because random.choice([]) will raise
        os.kill(random.choice(preemptible_workers), 15)
```

- The memory limits are at `400e6` times 4 workers now.  But I'm not sure if binder will be fine if we really use all of this at the same time.  (They're limiting to 1 or 2 GB depending on load, IIRC.)

willirath:
>  Add example dealing with Exceptions in user code? Relevant docs: http://distributed.dask.org/en/latest/resilience.html#user-code-failures

I think this deserves a separate example (could add more than one).

resilience.ipynb
```diff
@@ -0,0 +1,205 @@
...
+    "Relevant docs: <http://distributed.dask.org/en/latest/resilience.html#hardware-failures>"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Increase resilience\n",
+    "\n",
+    "Whenever a worker shuts down, the scheduler will increment the suspicousness counter of _all_ tasks that were assigned (not necessarily computing) to the worker in question.  Whenever the suspiciousness of a task exceeds a certain threshold (3 bu default), the task will be considered broken.  We want to compute many tasks on only a few workers with workers shutting down randomly.  So we expect the suspiciousness of all tasks to grow rapidly.  Let's effectively disable the threshold by setting it to a large number:"
```
TomAugspurger: bu -> by

resilience.ipynb
```diff
@@ -0,0 +1,205 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "import dask\n",
+    "\n",
+    "# needs to be done before importing anything from dask.distributed \n",
```
TomAugspurger: Is this true? Or does it just need to be done before creating a client / scheduler / workers?

resilience.ipynb
```diff
@@ -0,0 +1,205 @@
...
+    "Let's mark two worker process id's as non-preemptible."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "_all_current_workers = [w.pid for w in cluster.scheduler.workers.values()]\n",
```
TomAugspurger: Any reason this starts with an underscore?

resilience.ipynb
willirath: Yes.
<https://github.com/dask/distributed/blob/e846991d93054a29e28528224c63db69972ffc9c/distributed/scheduler.py#L76> sets the var `ALLOWED_FAILURES` at which will be used during initialization of the scheduler here <https://github.com/dask/distributed/blob/e846991d93054a29e28528224c63db69972ffc9c/distributed/scheduler.py#L832>.

resilience.ipynb
willirath: Thanks.  I'll change this.

resilience.ipynb
willirath: No idea, actually.  Will change this.

resilience.ipynb
willirath: I'm not really sure why this is done. Probably due to a policy of generally avoiding repeated `dask.config.get` calls for performance reasons?

TomAugspurger: It's probably unintentional. Schedulers aren't created that often. https://github.com/dask/distributed/pull/2761

willirath:
I've removed the comment.. Thanks for having a look, @TomAugspurger.

resilience.ipynb
```diff
@@ -0,0 +1,204 @@
...
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Resilience against hardware failures\n",
+    "\n",
+    "Scenario:  We have a cluster that partially consists of preemptible ressources.  That is, we'll have to deal with workers suddenly being shut down during computation.\n",
+    "\n",
+    "Relevant docs: <http://distributed.dask.org/en/latest/resilience.html#hardware-failures>"
```
guillaumeeb: Maybe we should give some examples : dask-jobqueue cluster with jobqueue system walltime, dask-kubernetes with preemptible resources?

resilience.ipynb
```diff
@@ -0,0 +1,204 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Increase resilience\n",
+    "\n",
+    "Whenever a worker shuts down, the scheduler will increment the suspicousness counter of _all_ tasks that were assigned (not necessarily computing) to the worker in question.  Whenever the suspiciousness of a task exceeds a certain threshold (3 by default), the task will be considered broken.  We want to compute many tasks on only a few workers with workers shutting down randomly.  So we expect the suspiciousness of all tasks to grow rapidly.  Let's effectively disable the threshold by setting it to a large number:"
```
guillaumeeb: Interesting! Isn't there another way? Add a per task counter in dask somehow?

TomAugspurger: If there are multiple tasks executing at once when worker is killed (say from a segfault) it may be hard to know exactly which task caused the segfault.

resilience.ipynb
willirath: Yes, we could mention this here.

willirath:
@guillaumeeb part of the reason for creating this example was to make others explore Dask's resilience. There's a lot of potential!

I agree that this should be expanded on in a jobqueue or kubernetes context. But I'd argue for keeping this notebook here short an simple.

resilience.ipynb
guillaumeeb: Hm, I agree, but shouldn't it count as one global failure? And we should be allowed 3 times this failure. As far as I understand, currently if there are more than 3 tasks failing due to this one task error, the computation is halted?

guillaumeeb:
And thanks for that @willirath! 

And yes, just keep this notebook simple, and elaborate more into dask-jobqueue context, That's one good example for a dedicated notebook gallery.

resilience.ipynb
willirath: Ah, now I understand where the misunderstanding is: This **_is_** a per task counter. But all tasks belonging to a given worker when this worker dies will be marked as suspicious. With a large number of tasks per worker, this counter still grows fast. There's no way for the scheduler to know which task was active at the time of failure. So all are marked.

(And it's also not always clear that the task currently computing is the problem. Think of a task that needs more memory or disk than the cluster can cope with. This task might not directly lead to failure but just let reasonably sized tasks trigger OOM kills further down the road.)

guillaumeeb: So you confirm there is one counter for each task? And these counters take a +1 failure for all the tasks that were on a given Worker when it dies? That means that by default, we should be robust to 3 workers down in one computation, is this the case?

resilience.ipynb
willirath: Done.

resilience.ipynb
```diff
@@ -0,0 +1,205 @@
...
```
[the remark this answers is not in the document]
willirath: Done.

resilience.ipynb
```diff
@@ -0,0 +1,205 @@
...
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "N = 5000\n",
+    "\n",
+    "x = db.from_sequence(range(N), npartitions=N // 10)\n",
+    "\n",
+    "mults = x.map(multiply_by_two)\n",
+    "\n",
+    "summed = mults.reduction(sum, sum)"
```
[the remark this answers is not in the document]
willirath: Done.

willirath:
I've added links to Dask Kubernetes and Dask Jobqueue.  And after making the example computation smaller and faster and letting fewer workers die, I've reduced the `allowed_failures` threshold to a still comfortable `100`.

TomAugspurger:
Restarted the failed build. LMK if you see that it fails / succeeds before I do.

TomAugspurger:
All green. Thanks @willirath!

willirath:
Let's revisit this as soon as distributed v2 is out.  With https://github.com/dask/distributed/pull/2761 and later modification of `allowed_failures` this will be easier to understand.

PR #76, opened 2019-06-09
Add stencils with Numba application notebook

mrocklin:


TomAugspurger:
Looks good. We're OK with adding numba to the default install (if it wasn't
pulled in implicitly)? It seems reasonable, but wanted to make sure.

mrocklin:
I think it's worth it.  Last time I checked it was something lime 40MB.  I suspect that it's general purpose enough that we'll find other uses for it.

PR #81, opened 2019-06-17
Add links to tutorials in some places

martindurant:


array.ipynb
```diff
@@ -130,6 +130,15 @@
...
     "%time y.sum().compute()"
    ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Further Reading \n",
+    "\n",
+    "A more in-depth guide to working with Dask arrays can be found in the [dask tutorial](https://github.com/dask/dask-tutorial), notebook 03."
```
mrocklin: We might want to avoid enumerating the notebook here.  I suspect that a reader will be able to find the array notebook, given the name.  I wouldn't be surprised if the notebook number changes in a year or so, and it would be nice not to have to update these references when that happens.

mrocklin: Should we also include the documentation for each section here?  That would also convert this text into a bulleted list, which I think tends to increase engagement a bit.

mrocklin:
A couple of tiny comments, but I'd also be very happy to just merge this in the interests of charging ahead.

Thanks @martindurant !

array.ipynb
martindurant: > We might want to avoid enumerating the notebook here

I agree, bu I see this as a medium-term only, until the tutorial is also (mostly) binderised, and then we can link directly to the notebook(s) in question.

mrocklin:
OK.  Sounds great.  Merging!

PR #83, opened 2019-06-22
Update for JLab rc

ian-r-rose:
Workspace loading is working again, now with less manually patching of the workspace file.


mrocklin:
This is great!  Thank you so much for tracking and resolving this @ian-r-rose

PR #89, opened 2019-07-18
Added array optimimzation fuse notebook

TomAugspurger:
From https://github.com/dask/dask/issues/5105.

### DIFF
--- a/applications/array-optimization.ipynb
+++ b/applications/array-optimization.ipynb
@@ -1,0 +1,231 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Array Scheduling Optimization\n",
+    "\n",
+    "In this example, we'll perform a typical workflow\n",
+    "\n",
+    "* Load batches of data\n",
+    "* Stack batches into a single Dask Array\n",
+    "* Rechunk the data for downstream processing\n",
+    "* Write rechunked data to disk, using Zarr\n",
+    "\n",
+    "Depending on the rechunking pattern, this is an embarassingly parallel opertion. However,\n",
+    "the Dask scheduler doesn't necessarily know that, which may result in sub-optimal scheduling. We'll configure the optimization settings to ensure that we have ideal scheduling to minimize data transfer across the cluster."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "from dask.distributed import Client\n",
+    "import dask.array as da\n",
+    "import zarr\n",
+    "\n",
+    "client = Client(processes=False, threads_per_worker=4,\n",
+    "                n_workers=1, memory_limit='2GB')\n",
+    "client"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Rather than loading data from disk, we'll generate random data in memory.\n",
+    "We'll have 25 batches, each of which has shape `(2,000,000 x 90,000)`.\n",
+    "\n",
+    "We rechunk along just the first axis (preserving the chunking on the second axis)."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "inputs = [da.random.random(size=2_000_000, chunks=90_000)\n",
+    "          for _ in range(25)]\n",
+    "inputs_stacked = da.vstack(inputs)\n",
+    "inputs_rechunked = inputs_stacked.rechunk((50, 90_000))"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "inputs_stacked"
+   ]
+  },
+  {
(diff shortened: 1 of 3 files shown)


applications/array-optimization.ipynb
```diff
@@ -0,0 +1,240 @@
...
+    "                n_workers=1, memory_limit='2GB')\n",
+    "client"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Rather than loading data from disk, we'll generate random data in memory.\n",
+    "We'll have 25 batches, each of which has shape `(2,000,000 x 90,000)`.\n",
```
alimanfoo:
```suggestion
    "We'll have 25 batches, each of which has shape `(2,000,000,)` and chunk shape `(90,000,)`.\n",
```

applications/array-optimization.ipynb
```diff
@@ -0,0 +1,240 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "inputs = [da.random.random(size=2_000_000, chunks=90_000)\n",
+    "          for _ in range(25)]\n",
```
alimanfoo: Minor point, but it might be good to increase the number of inputs from 25 to something greater than 50 (e.g., 100), and then also increase the fuse_ave_width to 50, so that the connection is clear between the value of fuse_ave_width and the size of the chunks in the rechunk operation (and not the number of inputs).

applications/array-optimization.ipynb
```diff
@@ -0,0 +1,240 @@
...
+    "Examining the structure of the stacking and rechunking, we can see that the problem is embarassingly parallel. We'll take a look at the task graph for two of the blocks."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "inputs_rechunked.blocks[0, :2].visualize(optimize_graph=True)"
```
alimanfoo: Nice trick, didn't know about this :-)

applications/array-optimization.ipynb
```diff
@@ -0,0 +1,240 @@
...
+    "\n",
+    "with dask.config.set(fuse_ave_width=25):\n",
+    "    write = inputs_rechunked.store(dest, lock=False)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Overall, we notice larger chunks, and no communcation. Each of these contributes to a faster runtime.\n",
```
alimanfoo:
```suggestion
    "Overall, we notice larger chunks, and no communication. Each of these contributes to a faster runtime.\n",
```

applications/array-optimization.ipynb
```diff
@@ -0,0 +1,240 @@
...
+    "\n",
+    "Fusing tasks is not an unambiguously good thing. As the docs for :ref:`dask.optimize.fuse` state\n",
+    "\n",
+    "> This trades parallelism opportunities for faster scheduling by making tasks less granular.\n",
+    "\n",
+    "In this case, *we* know that we can already acheive ideal parallelism by the coarse-grained embarassinly parallel nature of the problem. We're happy to have less fine-grained parllelism since we know we'll still be able to saturate the cluster."
```
alimanfoo:
```suggestion
    "In this case, *we* know that we can already achieve ideal parallelism by the coarse-grained embarrassingly parallel nature of the problem. We're happy to have less fine-grained parallelism since we know we'll still be able to saturate the cluster."
```

TomAugspurger:
Thanks @alimanfoo, I've applied your suggestions.

@mrocklin do you have high-level thoughts on this? Does this feel like we're just documenting a workaround to a weakness of Dask that we should instead be fixing?

mrocklin:
Yes, to me this notebook seems perhaps overly-specific to a single use
case.  I'm having trouble finding ways to generalize this notebook to other
situations.  I think that a general example of optimization would be
useful.  There are plenty of cases where this comes up, such as in ML
workloads where you really want X and y to be co-allocated.  That case
might also be a bit simpler.

mrocklin:
Although in general in many of these cases I think that we can improve them
just by expanding Blockwise and HighLevelGraph operator fusion out to data
access operations

martindurant:
@TomAugspurger , did you have plans to try to make the story here more general?

TomAugspurger:
Not at the moment.

TomAugspurger:
@mrocklin question on the HLG fusion: would you expect adding additional
operations to the end of a task graph (e.g. `.store`) to potentially result in
more fusion *earlier* on? My guess is that extra tasks won't lead to more fusion
earlier on, but I may be misreading `fuse`.

I ask because when I look at just the creation / stacking / rechunking, we don't
get fusion with the default parameters:

```python
import dask.array as da

inputs = [da.random.random(size=500_000, chunks=90_000)
          for _ in range(5)]
inputs_stacked = da.vstack(inputs)
inputs_rechunked = inputs_stacked.rechunk((50, 90_000))
inputs_rechunked.visualize(optimize_graph=True)
```

image


So unless adding a `.store()` to the end results in more fusion earlier on (in
the creation / stacking / rechunking phase), we won't be solving this use-case.

PR #91, opened 2019-08-01
Add user survey results.

TomAugspurger:
When compressed, the dataset is smaller than the notebook and our
image files, so I've included it in the repo.

Should we include the analysis I did for the blog post? Or should we leave it relatively spare? My preference is for including the blog post analysis.


mrocklin:
Maybe rename `results.csv.gz` to `dask-user-survey-2019.csv.gz`?

TomAugspurger:
Will do.

Want to publish this on Monday so we don't have two posts on the same day?

mrocklin:
Sure.

mrocklin:
I also haven't publicized the release post yet if you wanted to switch around the order.

TomAugspurger:
I think release first makes the most sense because it's already out.

TomAugspurger:
This and the blog post https://github.com/dask/dask-blog/pull/45 should be good to go. Merging later today.

We'll merge this a bit before the blog post, since the post references this notebook (on master).

PR #108, opened 2019-10-03
Add notebook on evolving workflows

mrocklin:
Fixes https://stackoverflow.com/questions/58209596/dask-parallelizing-across-nested-lists

cc @ericmjl


ericmjl:
Oooh thank you @mrocklin! :smile:

Issue #111, opened 2019-10-10
Use an already trained Torch model to predict on lots of data 


mrocklin:
Extending on #35 it would be nice to have an example using Dask and Torch together to parallelize prediction.  This should be a simple embarrassingly parallel use case, but I suspect that it would be pragmatic for lots of folks.  

The challenge, I think, is constructing a simple example that hopefully doesn't get too much into Torch or a dataset.  In my ideal world this would be something like 

```python
import torchvision
model = torchvision.get_model("model_name")

dataset = get_canned_dataset()
>>> imshow(dataset[0])  # show an example image

>>> model.predict(dataset[0])
"this is a cat"

... # then dask things here
```

Does anyone have good pointers to such a simple case? 

cc @stsievert @TomAugspurger @AlbertDeFusco

stsievert:
cc @muammar

In addition to this example, I'd also link to integration with a Scikit-learn wrapper for PyTorch [skorch] and Dask-ML's [ParallelPostFit][1].

[skorch]:https://skorch.readthedocs.io/en/stable/
[1]:https://ml.dask.org/modules/generated/dask_ml.wrappers.ParallelPostFit.html

TomAugspurger:
Should have an example ready tomorrow.

> hopefully doesn't get too much into Torch or a dataset.

I think we'll want to go into some detail about `torch.utils.data.Dataset`, because it's not 100% straightforward how to get the data loaded onto workers. To predict for a directory of images, I had to write the following myself

```python
import glob

from PIL import Image


def default_loader(path, fs=__builtins__):
    with fs.open(path, 'rb') as f:
        img = Image.open(f).convert("RGB")
        return img


class FileDataset(torch.utils.data.Dataset):
    def __init__(self, files, transform=None, target_transform=None,
                 classes=None,
                 loader=default_loader):
        self.files = files
        self.transform = transform
        self.target_transform = target_transform
        self.loader = loader
        if classes is None:
            classes = list(sorted(set(x.split(os.path.sep)[-2] for x in files)))
        else:
            classes = list(classes)
        self.classes = classes

    def __len__(self):
        return len(self.files)
    
    def __getitem__(self, index):
        filename = self.files[index]
        img = self.loader(filename)
        target = self.classes.index(filename.split(os.path.sep)[-2])
        
        if self.transform is not None:
            img = self.transform(img)
            
        if self.target_transform is not None:
            target = self.target_transform(target)
        
        return img, target
```

and use it as

```python
[...]

PR #114, opened 2019-10-16
Add torch prediction example

TomAugspurger:
Closes https://github.com/dask/dask-examples/issues/111.

Because pytorch / torchvision are so large, I don't think it's feasible to include them in the base image. 

I've opted to exclude this notebook from execution and have instead included the rendered output in the source.


mrocklin:
Thanks for doing this Tom! 

I wonder if we can do everything in a single compute call. Currently it looks like we use dask.delayed to construct batches, and then we make a function that calls dask.compute on a *single* batch, and then uses torch.  Then we map that function across all of the batches.  This seems to mix things in a way that, while it works here, I suspect might trip up users in the future.

Instead, I wonder if we might do something like this:

```python
images = [load(fn) for fn in ...]
batches = [...]
model = dask.delayed(finetune_model)()

@dask.delayed
def predict(batch, model):
    with torch.no_grad():
        out = model(batch)
        _, predicted = torch.max(out, 1)
        predicted = predicted.numpy()
    return predicted

predictions = [predict(batch, model) for batch in batches]

predictions = dask.compute(*predictions)
```

TomAugspurger:
Yeah, that should be doable, I'll just need to restructure the narrative a
bit. It did feel strange mixing delayed and Future.

TomAugspurger:
Implemented essentially that, aside from `model = dask.delayed(finetune_model)()`. PyTorch didn't like the training ending up on a Worker (something about a subprocess daemon).

So I train the model on the client, and use

```
delayed_model = dask.delayed(model)
```

as the object passed to the delayed `predict`.

mrocklin:
Some other small comments

- We might want to remove the `def batch` function
- We shouldn't need the `dask.compute` call in the `predict` function
-  Typo with `neurel`
-  "For a larger dataset, you would want to build up a dask.Array using dask.Array.from_delayed, or write the data to a shared filesystem or cloud storage."  I think that array is orthogonal here.  If it was a larger dataset then they might want to use persist or, as you suggest write things.  Array itself should be no different than delayed though when it comes to handling large outputs.

TomAugspurger:
> For a larger dataset, you would want to build up a dask.Array

That was unclear. I meant you might want to build up an array from the `predictions`, rather than `dask.compute(...)`. No change to any previous step.

TomAugspurger:
Planning to merge this later today.

mrocklin:
Two comments:

1. In other cases where we have depended on large packages that we didn't want to include in the binder environment we call `!pip install foo` on the first cell.  Would it makes sense to do this here or is PyTorch too big?
2.  We'll need to be a little bit careful with section headers.  The section headers in the notebook end up populating the table of contents.  In particular, we should have a top-level / h1 section header at the top of the document with the title of the document like `# Scale Pytorch Predictions` or something similar that matches the form of the other titles (see the TOC on https://examples.dask.org).  We should use `h2` and other headers appropriately so that we get the right nesting.

mrocklin:
The content here seems great to me

TomAugspurger:
Fixed the header, thanks.

Will try running the example. The previous build (not running) took 21 minutes.

TomAugspurger:
I'm not sure why the `torchvision` import is failing.

TomAugspurger:
Status update: when I changed to `!pip install pytorch torchvision` (or `!conda`) the CI was failing in an `import torchvision`. I wasn't able to figure out why.

If others are OK with it, I'd prefer to just merge this as is (with the output included, and the execution skipped).

mrocklin:
Sounds good to me

PR #115, opened 2019-10-16
Delete unused file

jsignell:
I am fairly confident that this file isn't actually being used. It would be good to trigger a dev build though, just in case.


quasiben:
I don't think it's being used either.  @TomAugspurger i want to double check with you

quasiben:
@jsignell this is your first contribution.  welcome!

jsignell:
Yes my goal is to have only red contributions :)

jrbourbeau:
Thanks @jsignell! 

> Yes my goal is to have only red contributions :)

+1

PR #130, opened 2020-01-12
Fix CI

jrbourbeau:


jrbourbeau:
This PR:

- Adds `prefect` to the conda environment used (not sure why this was passing CI before)

- Updates `dask-ml` to the latest release (1.2.0)

mrocklin:
Some of the notebooks will install their dependencies in the first cell

```
!pip install prefect[viz]
```

This is to avoid bloating out the binder environment.  However, if these dependencies are small enough then it probably makes sense to include them in the environment.yaml file as you've done here.  If we do that then we should probably remove the pip install line.

jrbourbeau:
Closing this PR as CI is now passing on `master`

PR #139, opened 2020-03-06
Pin JupyterLab to 1.x and add repo2docker CI build

jrbourbeau:
We're temporarily pinning JupyterLab and the dask lab extension to avoid issues with other JL extensions that aren't compatible with JupyterLab 2.0 yet. 

I've also added a GitHub Actions workflow to our CI that runs `repo2docker` on the tutorial to test that binder deployments will work

We did the same thing over in dask-tutorial https://github.com/dask/dask-tutorial/pull/166


TomAugspurger:
Thanks!

PR #143, opened 2020-04-10
Add timeseries forecasting with Prophet example

TomAugspurger:
This example shows how to parallelize the model diagnostics stage of time series forecasting with Prophet.

As written, I think it relies on the master version of Prophet. I'll see if I can make it work with the currently released version. Mostly just need a thing to link to right now.


TomAugspurger:
The CI failure is from the released prophet version not supporting the keyword we need to easily do parallel prediction. Attempting to do things with the released version will be a bit ugly, so I'd prefer to wait for a prophet release.

In the meantime we can just let this sit here, or we can skip execution of this notebook and add a caveat saying it depends on an unreleased Prophet. Or we could try installing Prophet with pip from GitHub, but I don't recall how to do that when the top-level of the package doesn't have a setup.py :)

mrocklin:
Cool stuff.  Thanks for doing this @TomAugspurger .  I'm curious, was this motivated by anything in particular, or did you just read a paper and decide that it would be easy to do?

How was performance?  Things scaled well I suppose?  Any GIL issues with Prophet?

TomAugspurger:
A twitter conversation between @jsebold and Sean Taylor reminded me that I wanted to look into this. And I've been extremely unproductive at other things lately :)

>  Any GIL issues with Prophet?

Here's a performance report from my local machine. Going to try it out on a cluster in a bit https://gistcdn.rawgit.org/TomAugspurger/9c7bfcdbb8ebc2f502c30713e2b38ab3/9a40a44f8c1c551fd13f322b14056cdf1409fd9e/dask-report.html

mrocklin:
> And I've been extremely unproductive at other things lately :)

Could have fooled me :)

TomAugspurger:
This should be passing now. Installing Prophet from source inside the notebook adds ~2.5 minutes to the build (20 minutes -> 22:30 minutes), which I think is acceptable for now.

TomAugspurger:
Status update: I opened https://github.com/facebook/prophet/pull/1434 to make this a bit smoother from the user's end. We'll see where it goes before merging this.

TomAugspurger:
https://github.com/facebook/prophet/pull/1434 is in prophet master. I've updated the example to use `parallel="dask"`. I think this can be merged anytime, and we can update the example to install from PyPI rather than prophet master once there's a release.

TomAugspurger:
> Any GIL issues with Prophet?

There definitely are GIL issues. Prophet does lots of stuff with pandas DataFrames. At a glance it looks like a decent chunk can be done on just ndarrays, but it'll be a while before I can look more.

mrocklin:
Woo!

Tweet?

TomAugspurger:
Yeah, I'll take care of that.

TomAugspurger:
Seems like this won't really be runnable on binder. The prophet install step is taking quite a bit longer on their instances than it did on travis.

PR #145, opened 2020-05-06
Blockwise metaestimator example

TomAugspurger:
Rendered at https://nbviewer.jupyter.org/gist/TomAugspurger/88ea870feb550e481652bbe701e5c750

### DIFF
--- a/binder/environment.yml
+++ b/binder/environment.yml
@@ -5,7 +5,7 @@ dependencies:
   - bokeh=1.4.0
   - dask=2.9.1
   - dask-image=0.2.0
-  - dask-ml=1.2.0
+  - dask-ml=1.4.0
   # TODO: Bump dask-labextension to >=2 once extensions have been made compatibility updates
   - dask-labextension=1.1.0
   # TODO: Bump jupyterlab to 2.x once extensions have been made compatibility updates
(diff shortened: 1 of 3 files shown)


mrocklin:
This seems great, and also really pragmatic.  I'm curious, was there a specific problem that motivated this work?

> The data passed to post-fit methods like .predict, .score, etc. does not need the same partitioning as the training data. To emphasize this, let's filter some rows from the original dataset and repartition it.

Putting on my naive user hat I read this as "Oh, I didn't realize that that would be a problem, but I'm glad that it isn't".  Given that, it might be useful to remove it, or perhaps reduce it to a note so that devs can be made aware, but most users don't get derailed by it.

TomAugspurger:
Moved the note about partitioning to a small aside. That was included to emphasize that each estimator in the ensemble is used for each prediction, rather than something like "estimator 1 predicts for partition 1, estimator 2 predicts for partition 2". Agreed that the previous version over-emphasized it.

This came out of a discussion with @nbren12. IIRC they'd toyed around with something like this internally and had some success.

mrocklin:
> This came out of a discussion with @nbren12. IIRC they'd toyed around with something like this internally and had some success.

Great!  This seems like an easy win.  I'm glad to hear that the method has had practical utility.

TomAugspurger:
Thanks for the feedback!

PR #153, opened 2020-06-16
Fixing Pytorch batch Prediction

anaszain89:
Fixed the error when stacking the batches
Moved the model to the CPU to run inference. This avoids error if fine-tuning was performed on the GPU


TomAugspurger:
Thanks!

PR #157, opened 2020-07-11
Add example notebook for messy csv data

dharhas:
This example shows how to read a collection of csv files into a dask dataframe when some of the data files are corrupted. It uses delayed and dd.from_delayed

### DIFF
--- a/dataframes/04-reading-messy-data-into-dataframes.ipynb
+++ b/dataframes/04-reading-messy-data-into-dataframes.ipynb
@@ -224,7 +224,11 @@
   {
    "cell_type": "code",
    "execution_count": null,
-   "metadata": {},
+   "metadata": {
+    "tags": [
+     "raises-exception"
+    ]
+   },
    "outputs": [],
    "source": [
     "df.describe().compute()"
@@ -255,17 +259,18 @@
    "outputs": [],
    "source": [
     "import numpy as np\n",
+    "import io\n",
     "\n",
     "def read_data(filename):\n",
     "    \n",
-    "    # for this to work we need to explicitly set the datatypes of our dataframe \n",
-    "    dtypes = np.dtype([('id', int), ('name', str), ('x', float), ('y', float)])\n",
+    "    # for this to work we need to explicitly set the datatypes of our pandas dataframe \n",
+    "    dtypes = {'id': int, 'name': str, 'x': float, 'y': float}\n",
     "    try:\n",
     "        # try reading in the data with pandas \n",
     "        df = pd.read_csv(filename, dtype=dtypes)\n",
     "    except:\n",
-    "        # if this fails create an empty dataframe with the same dtypes as the good data\n",
-    "        df = pd.DataFrame(np.empty(0, dtype=dtypes))\n",
+    "        # if this fails create an empty pandas dataframe with the same dtypes as the good data\n",
+    "        df = pd.read_csv(io.StringIO(''), names=dtypes.keys(), dtype=dtypes)\n",
     "    \n",
     "    # for the case with the missing column, add a column of data with NaN's\n",
     "    if 'y' not in df.columns:\n",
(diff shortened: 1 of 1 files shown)


dataframes/04-reading-messy-data-into-dataframes.ipynb
```diff
@@ -0,0 +1,502 @@
+{
```
jcrist, line 1: <p>I didn't know that <code>dtypes</code>  could be a <code>np.dtype</code>  object. Since not all pandas dtypes have numpy counterparts, this should be a <code>dict</code>  (e.g. <code>{"id": int, "name": str, ...}</code> ) instead IMO.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/dask/dask-examples/pull/157/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='24'/>

jcrist, line 1: <p>I don't think this note is necessary here, I find it a bit distracting.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/dask/dask-examples/pull/157/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='31'/>

jcrist, line 1: <p>This will sneakily read the first file to infer the schema if you don't pass in <code>meta=...</code> . Depending on how expensive computing the first partition is this could be fine, or it might be bad.</p><p> </p><p>Here the <code>meta</code>  would be something like:</p><p> </p><pre class="ql-syntax" spellcheck="false">meta = [("id", int), ("name", str), ("x", float), ("y", float)]
</pre><p> </p><p>A dict from column name -&gt; dtype also works (e.g. <code>{"id": int, ...}</code>).  This is similar enough to the <code>dtype</code>  kwarg used above that you might share them.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/dask/dask-examples/pull/157/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='41'/>

jcrist, line 1: <p>Perhaps add a conclusion paragraph recapping what was done (using <code>dask.delayed</code>  and <code>dask.dataframe.from_delayed</code>  to create a custom reader function that handles messy data). The end here feels a bit abrupt.</p><p> </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/dask/dask-examples/pull/157/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='45'/>

jcrist, line 1: <p>To make this pass our CI run, any cell that intentionally errors needs to be tagged with a <code>raises-exception </code> tag. See https://nbsphinx.readthedocs.io/en/0.4.3/allow-errors-per-cell.html for more information.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/dask/dask-examples/pull/157/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='21'/>

dharhas: yeah so I had an issue that it looks like the pd.DataFrame constructor cannot take a compound datatype. So the only other option are to use either:

```python
import io
dtypes={"id": int, "name": str, "x": float, "y": float}
pd.read_csv(io.StringIO(""), names=dtypes.keys(), dtype=dtypes)
```

or 

```python
df = pd.DataFrame({'id':pd.Series([], dtype=str),
                   'name':pd.Series([], dtype=int),
                   'x':pd.Series([], dtype=float),
                   'y':pd.Series([], dtype=float)})
```

Do you prefer either one of these? I think I would go with the first option.

dharhas: <p>I originally had the meta in there, will put it back in.</p>

dharhas: <p>good idea</p>

jcrist: <p>Ah, didn't know that. The first one looks best to me.</p>

jcrist:
LGTM! Thanks, merging.

PR #160, opened 2020-07-27
CountVectorizer example

TomAugspurger:


mrocklin:
There is already this:
https://examples.dask.org/machine-learning/text-vectorization.html

Maybe we can roll this into that somehow?

TomAugspurger:
Merged them into a single "Working with text data" notebook that starts with different comparing different vectorizers (HashingVectorizer, CountVectorizer) and ends with the full pipeline.

machine-learning/text-count-vectorizer.ipynb
```diff
@@ -0,0 +1,212 @@
...
+    "partition_size = len(news['data']) // npartitions\n",
+    "\n",
+    "lengths = np.cumsum([partition_size] * npartitions)\n",
+    "lengths = [0] + list(lengths) + [None]\n",
+    "\n",
+    "slices = [slice(a, b) for a, b in\n",
+    "          toolz.sliding_window(2, lengths)]\n",
+    "# Notice the persist here! More details later.\n",
+    "documents = db.from_delayed([load_news(x) for x in slices]).persist()\n",
+    "documents"
```
mrocklin: We could also call `db.read_sequence(..., npartitions=10).persist()` and then call `client.rebalance()`

Given that people are going to blindly copy-paste whatever we do anyway I'd personally rather that they see this.  It's a bit more in line with ordinary behavior I think.

machine-learning/text-count-vectorizer.ipynb
```diff
@@ -0,0 +1,212 @@
...
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "import dask_ml.feature_extraction.text"
+   ]
+  },
+  {
+   "cell_type": "code",
```
mrocklin: I recommend merging adjacent code cells, if only to cut down on Ctrl-Enter pressing.

machine-learning/text-count-vectorizer.ipynb
```diff
@@ -0,0 +1,212 @@
...
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "vocabulary = vectorizer.vocabulary_\n",
+    "remote_vocabulary, = client.scatter([vocabulary], broadcast=True)\n",
+    "\n",
+    "vectorizer2 = dask_ml.feature_extraction.text.CountVectorizer(\n",
+    "    vocabulary=remote_vocabulary\n",
+    ")"
```
mrocklin: Is there a well defined vocabulary that we can use somewhere?  Maybe in `nltk`?  I'm concerned that people will see this, and think that they should copy the vocabulary off of one CountVectorizer and then pass it to another.  

Also, do we need the scatter? Can you verify that if vocabulary is included directly in the `vocabulary=` keyword argument that it will occupy only a single task, and not be in many of them?

TomAugspurger: I'm not sure about nltk, but probably not worth adding it to the environment just for this example. I noted that you'd probably get this from an external source in practice.

I think that as of https://github.com/dask/dask-ml/pull/719, the answer to your question about user-provided vocabulary being in one task is "yes". But that change hasn't been released yet.

PR #166, opened 2020-09-01
Fix typos and headings on dask.org

TAdeJong:
I noticed still a few "iamge" typo's remained, so fixed these. Also, any markdown heading without a newline at the end seems to not render at examples.dask.org:
image

This attempts to fix that by adding newlines to each heading not having one.


TAdeJong:
@jrbourbeau : Unfortunately, it does seem to be some further weird thing with `nbsphinx`, as the typo's are fixed on the website, but the headings still don't appear correctly, while they do appear correctly in `binder` and in the github static renderer.

PR #171, opened 2020-10-29
Added an example notebook on dask-sql.

### LINKED ISSUE #170: Add dask-sql example(s)
As was commented on the dask-sql repo

https://github.com/nils-braun/dask-sql/pull/60#issuecomment-717541194

nils-braun:
Fixes #170
This is a first draft for an example notebook showcasing dask-sql.
It is not very long and only shows the main features (e.g. no custom functions, no reusing results as new tables etc) - but for a short overview I think it is fine.
If you would love to see more content or anything different, I am happy to improve!


nils-braun:
I am a bit confused. The installation works with `mamba`, but now with `conda` - which makes no sense... With conda it says that "nodejs" and "numpy" are missing. Do you think it is fine to use mamba here (also, it is much faster) - even though it might be less known?

jrbourbeau:
Woo, thanks for adding this @nils-braun! 

Re: installing `dask-sql`, could you try adding it to ` dask-examples/binder/environment.yml ` to see if we can successfully solve with that added dependency? That might be the path of least resistance

mrocklin:
I'm glad to see this.  Some comments/questions:

1.  Should we rename this to `sql.ipynb` ?
2.  Should we put it in the root folder rather than applications?
3.  We should add it to the index.rst file
4.  How much does conda installing this increase the size of the docker image?  How large are the dependencies (I imagine calcicte might be large)

nils-braun:
Very reasonable comments and questions @mrocklin and @jrbourbeau  - thanks for that.
1. I have done so.
2. I am undecided. dask-sql is not part of the main dask repo, but it could also be a main entry point for many users. I have done so in the newest changes.
3. Good point, have done so
4. Before installing dask-sqlm the conda env (/srv/conda/envs/notebook) has a size of 2.8G. After that, it is 3.2 GB. So around 400 MB. I guess the largest part of that is the JVM.
I have tested to add it to the environment file, lets see what the test says.

nils-braun:
So according to the tests, the package solving was successful.
However, now there is an error popping up in the "bag.ipynb" notebook - which (at least to me) seems to be unrelated and more a problem with pandas (although I could not reproduce the same error on my laptop).

mrocklin:
Right, I agree that those errors are probably unrelated.  

Hrm, 400MB for this seems large.  If possible I'd prefer to keep the installation in the notebook itself and outside of environment.yml (large images slow down container start times).

nils-braun:
Unfortunately, I am currently bugged by #173 - but apart from that, I think it should work now.

nils-braun:
I have updated the PR for the newest dask-sql version and also moved again to mamba for installation .
I tested the notebook on binder, lets see if the unittests work this time.

jacobtomlinson:
Closing in favour of #209

PR #174, opened 2020-11-04
Migrate to GitHub Actions

### LINKED ISSUE #172: Migrate CI to GitHub Actions
Due to changes in the Travis CI billing, the Dask org is migrating CI to GitHub Actions.

This repo contains a `.travis.yml` file which needs to be replaced with an equivalent `.github/workflows/ci.yml` file.

See dask/community#107 for more details.

jacobtomlinson:
Closes #172

Migrate the build and deploy job to GitHub Actions.

Builds will be done on pushes and PRs for CI purposes but deploy to `gh-pages` will only happen on master branch.


.github/workflows/ci-build.yaml
```diff
@@ -0,0 +1,48 @@
...
+  build-and-deploy:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout source
+        uses: actions/checkout@v2
+
+      - name: Install graphviz
+        shell: bash -l {0}
+        run: |
+          sudo apt-get update && sudo apt-get install graphviz
```
jrbourbeau, line 14: I think we can drop this as the `python-graphviz` conda package should install the system graphviz library

.github/workflows/ci-build.yaml
```diff
@@ -0,0 +1,48 @@
...
+
+      - name: Install SSH Client
+        if: ${{ github.ref == 'refs/heads/master' && github.event_name != 'pull_request'}}
+        uses: webfactory/ssh-agent@v0.4.1
+        with:
+          ssh-private-key: ${{ secrets.DEPLOY_KEY }}
+
+      - name: Deploy
+        if: ${{ github.ref == 'refs/heads/master' && github.event_name != 'pull_request'}}
+        uses: JamesIves/github-pages-deploy-action@3.7.1
```
jrbourbeau, line 43: It's not critical, but I wanted to point out [`peaceiris/actions-gh-pages`](https://github.com/peaceiris/actions-gh-pages) as another possible action we could use for deploying docs. It uses the built in `secrets.GITHUB_TOKEN` auth token and so doesn't require an additional deploy key got setting up an ssh client. That said, I'm happy to defer to you here

jrbourbeau (commented):
Woo, thanks @jacobtomlinson! This is great to see. I've left a couple of small comments, otherwise this LGTM

.github/workflows/ci-build.yaml
jacobtomlinson: Awesome!

.github/workflows/ci-build.yaml
jacobtomlinson: Hmm interesting. The one I've used here also supports that, but seems to advise in the docs that this can break with some repo configs and that a deploy key is more robust.

Happy to switch to the `GITHUB_TOKEN` approach to see if it works.

jacobtomlinson:
Thanks for the review @jrbourbeau!

I've implemented your suggestions and I'm going to hit merge as that's the only way we can really test the deploy works.

jacobtomlinson:
Success 🚀

image

jrbourbeau:
Woo! Thanks @jacobtomlinson!

jacobtomlinson:
I've just wrapped this up by removing the `DEPLOY_KEY` secret that I created and cleaning up the old deploy keys as they will no longer be needed.

Issue #152, opened 2020-06-04
Regularly check for broken links


hammer:
Inspired by https://github.com/dask/dask-examples/pull/151, which I encountered manually, I just ran http://examples.dask.org through a dead link checker and found a handful of broken links. I don't have time to fix them all right now but I just thought I'd drop the results here.

It might be a good idea to include a dead link check as part of the website deploy, but that may also be overkill!

| Status | URL | Source link text |
|-----------|-------|--------|
| -1 Invalid URL | <http://127.0.0.1:8787/status> | [http://127.0.0.1:8787/status](https://examples.dask.org/bag.html "http://127.0.0.1:8787/status") |
| 404 Not Found | <https://docs.dask.org/en/latest/bag-overview.html> | [Dask Bag Documentation](https://examples.dask.org/bag.html "Dask Bag Documentation") |
| 404 Not Found | <https://www.continuum.io/sites/default/files/dask_stacked.png> | [<No Text>](https://examples.dask.org/machine-learning/incremental.html) |
| -1 Invalid URL | <http://10.20.0.141:8787/status> | [http://10.20.0.141:8787/status](https://examples.dask.org/array.html "http://10.20.0.141:8787/status") |
| 404 Not Found | <https://ml.dask.org/examples/xgboost.html> | [http://ml.dask.org/examples/xgboost.html](https://examples.dask.org/machine-learning/xgboost.html "http://ml.dask.org/examples/xgboost.html") |
| 404 Not Found | <https://xgboost.readthedocs.io/en/latest/python/python_intro> | [https://xgboost.readthedocs.io/en/latest/python/python_intro](https://examples.dask.org/machine-learning/xgboost.html "https://xgboost.readthedocs.io/en/latest/python/python_intro") |
| 404 Not Found | <https://distributed.readthedocs.io/en/latest/local-cluster.html> | [local cluster](https://examples.dask.org/machine-learning/voting-classifier.html "local cluster") |
| 404 Not Found | <https://docs.scipy.org/doc/numpy-1.16.0/reference/c-api> | [https://docs.scipy.org/doc/numpy-1.16.0/reference/c-api](https://examples.dask.org/applications/stencils-with-numba.html "https://docs.scipy.org/doc/numpy-1.16.0/reference/c-api") |
[...]

PR #176, opened 2020-11-08
Fixes broken links (#152)

pratyakshajha:
Fix broken links listed in issue #152. I used pytest-check-links to automatically find broken links and fixed them as well. 

Draft request currently as part of 152. I will need to investigate on configuring Travis CI support for this. I am assuming adding the pytest checklinks command to the build should work. I'll try to add CI support and update this PR.


mrocklin:
Thank you @pratyakshajha 

Many of these fixes look good.  Some of them we may not want.  For example I recommend continuing to link to `/latest` docs.  Also, I we need to keep `../proxy/..` links.  They are helpful for people who run these notebooks live.

pratyakshajha:
Thanks @mrocklin. 
I will fix the links to pint to `/latest` doc links.
I was unsure about what to do with `../proxy/..` links but this makes more sense. I will fix those. 

Also, Pytest breaks for these `../proxy/..` links. I have tried to find some way to exclude it by some link patterns, but haven't found anything useful. Sphinx has an option to exclude some URL patterns while checking links but it can be slow, needs all notebooks running without errors (I had some trouble with fbprophet notebook) and lesser coverage as it can hit rate limits for some websites like Github. 
Sphinx can be used instead of Pytest if we really want to integrate link checking in the CI pipeline. Or, as an alternative, Pytest can be used with a suggestion to run it maunally in `CONTRIBUTING.md` or maybe somewhere contributors are aware about it. I will need some suggestion on how handle this and if this is needed as a part of this PR.

PR #180, opened 2021-01-01
Update notebook links

### LINKED ISSUE #169: Links in Image Processing page aren't working
The links at the top of the Image Processing page aren't working. Clicking any of the links at the top doesn't bring you down the page to the content as it should.
https://examples.dask.org/applications/image-processing.html

I'm thinking maybe it has to do with the extra apostrophe in the last (cleanup) link? Didn't have time to test that theory though.

FabioRosado:
Hello, this is my first time updating notebooks, I'm not entirely sure if I did it the right way - I've noticed that the python version was bumped to 3.8 so I might need to revert those changes manually.

This PR aims to fix #169 I've updated all the content links and tested them manually - they work in the notebooks and should work on the docs as well.

I've also noticed that the [ETL Pipelines with Prefect](https://examples.dask.org/applications/prefect-etl.html), on the Goals part, the code highlight seems broken, I've tried to put the `"""` on the same line as the rest of the code to see if that will fix it.

Let me know if you would like me to change something 😄 👍

### DIFF
--- a/applications/image-processing.ipynb
+++ b/applications/image-processing.ipynb
@@ -19,32 +19,31 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "1. [Setting up your environment](#environment_setup)\n",
-    "1. [Importing dask-image](#imports)\n",
-    "1. [Getting the example data](#example_data)\n",
-    "1. [Reading in image data](#reading_image_data)\n",
-    "    1. [Reading a single image](#reading_a_single_image)\n",
-    "    1. [Reading multiple images](#reading_multiple_images)\n",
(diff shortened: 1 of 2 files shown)


mrocklin:
>  I've noticed that the python version was bumped to 3.8 so I might need to revert those changes manually.

I wouldn't worry too much about this. Any time someone touches these files these revert to whatever version of Python that person was running. 

In general this looks good to me.  I've tried things out locally and they seem good. Let's merge this in, let CI do it's work, and check to see if everything is fine when rendered live. 

Thanks @FabioRosado !  Also, I notice that this is your first code contribution to this repository.  Welcome!

PR #181, opened 2021-01-22
Update futures.ipynb 

alex-rakowski:
The dynamically and statically assigned graphs resulted in different summation values. Changed line 321 `while seq.count() > 2:  # at least two futures left` to `while seq.count() > 1:  # at least two futures left`.  They now output the same value.


jacobtomlinson:
Thanks!

PR #190, opened 2021-05-11
Fix conda solve issue

jrbourbeau:
Avoid specifying the Python version in multiple places


jacobtomlinson:
Thanks for starting this PR @jrbourbeau.

I've pushed a couple more commits to try and unblock the CI:
- Looks like there is a dependency issue in repo2docker builds due to upstream changes in the `docker` package so I explicitly added the missing deps (jupyterhub/repo2docker/issues/1065)
- The notebooks were complaining about `prophet` not being installed to added that to the environment.

PR #205, opened 2021-12-27
add join example

kori73:
Added a join example as discussed in [dask](https://github.com/dask/dask/issues/6593).


jacobtomlinson (approved):
Seems reasonable to me. Many thanks.

Issue #206, opened 2022-01-31
Update dependencies and ensure all notebooks are working


bryanwweber:
Hi dask-examples maintainers! As I'm learning Dask after starting @coiled I thought about adding an example here, or at least doing something that could work in that direction. While getting set up, I noticed that a number of the dependencies of the repo are out-of-date or not available for the `aarch64`/`arm64` (Mac M1) platform. Specifically:

* `dask` is pinned to 2.20.0, but 2022.1.1 was released last week
* `dask-image` is pinned to 0.2.0 but 2021.12.0 is availble
* `dask-ml` is pinned to 1.6.0 but 2022.1.22 is available
* `python` is pinned to 3.8 (perhaps because of `pystan` below?)
* `py-xgboost` and `dask-xgboost` are either deprecated or not available
* prophet requires pystan==2.19.1.1, but pystan 3.3.0 is released (this is a known issue with prophet)
* pystan, httpstan, and simdjson are not available for `aarch64` via conda or PyPI
* `scikit-learn` is pinned to 0.23
* `scikit-image` is installed via pip in the example notebook

I'd like to propose updating all these dependencies and putting them into `binder/environment.yml`. Before I get in too deep, especially with `prophet` which is likely to be the most complicated one, I'd like some feedback on whether this is desirable and which versions of dependencies (especially `dask-*` libraries) have known conflicts or requirements.

On a related, but separate, note, I've been experimenting with creating a [`devcontainer`](https://code.visualstudio.com/docs/remote/containers) for this repo so that developers can run a full build of the site as they're working. Essentially, this is an add-on for VS Code that mounts a local directory into a pre-configured Docker container that is started by VS Code. All commands and terminals are run inside the container, so it can more-or-less replicate the environment on GH Actions to build the examples via Sphinx and serve them locally.

I think this would be a nice helper for anyone who wanted to add a new example, but it would require committing VS Code specific configuration to the repo. Feedback on this idea is welcome as well! Thanks :smiley:

bryanwweber:
> * prophet requires pystan==2.19.1.1, but pystan 3.3.0 is released (this is a known issue with prophet)
> * pystan, httpstan, and simdjson are not available for `aarch64` via conda or PyPI

[...]

Issue #173, opened 2020-11-03
Can not use "conda install in binder environment"


nils-braun:
It seems that to whatever reasons, it is not possible to install a package into the binder environment via conda.
I tried via the terminal or directly from within the jupyter notebook. I started binder by clicking one of the links from the dask-example page.

It does not really matter which package (I tried dask-sql, but also e.g. pandas, which is already installed). Both fail with:
```
Collecting package metadata (current_repodata.json): done
Solving environment: failed with initial frozen solve. Retrying with flexible solve.
Collecting package metadata (repodata.json): done
Solving environment: failed with initial frozen solve. Retrying with flexible solve.

PackagesNotFoundError: The following packages are missing from the target environment:
  - nodejs==14
  - numpy==1.18
```

The installation seems to work with mamba, which makes me think it is some problem in the way conda resolves the packages - but I have no idea what is going wild her. 

Now comes the "funny" part: If I explicitly install 
```
conda install numpy=1.18 nodejs=14
```
it actually works, even though conda list already shows me them as installed. It will downgrade some packages:

```
The following packages will be UPDATED:

  certifi                          2020.6.20-py38h32f6830_0 --> 2020.6.20-py38h924ce5b_2

The following packages will be DOWNGRADED:

  jupyterlab                                     2.1.5-py_0 --> 2.1.0-py_1
  pandas                               1.0.5-py38hcb8c335_0 --> 1.0.0-py38hb3f55d8_0
  scikit-learn                        0.23.2-py38h5d63f67_2 --> 0.23.0-py38h3a94b23_0
```

After that, I can also install other packages...

mrocklin:
cc'ing conda/dask experts @jakirkham @mariusvniekerk in case they have any insight

quasiben:
Can you get the conda version being used in the binder example ? If it's less than 4.9 we should probably upgrade it

nils-braun:
Thanks for your comment @quasiben. I just checked: The currently running image includes conda in version 4.8.2. If 4.9 will really help, that could be an issue.
Where is the conda version specified? Isn't this part of the mybinder installation process where we do not have a large number of configuration possibilities?

quasiben:
No that's probably not it then.  I launched an instanced on binder and things pleasantly worked with `mamba`. i would suggest either using mamba or pip, or, why not add directly to the env.yml file ?  Any reason not to do this @mrocklin  ?

nils-braun:
[...]

PR #207, opened 2022-02-02
Minimal update of deps to allow installing on ARM Mac

saulshanabrook:
This PR updates two of the dependencies in the `environment.yml` so that it is installable on an ARM mac. The previous pins had no ARM mac builds available on conda forge. I chose the oldest version that had an ARM build, to be minimally disruptive.

I am not sure if this is exactly the same issue as https://github.com/dask/dask-examples/issues/173 but is possibly overlapping. 

Also partially addresses updating all deps, as raised in https://github.com/dask/dask-examples/issues/206.

### DIFF
--- a/binder/environment.yml
+++ b/binder/environment.yml
@@ -3,13 +3,13 @@ channels:
   - conda-forge
 dependencies:
   - python=3.8
-  - bokeh=2.1.1
+  - bokeh=2.2.3
   - dask=2.20.0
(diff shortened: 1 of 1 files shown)


jrbourbeau (approved):
Thanks @saulshanabrook! Good to see you here : )

saulshanabrook:
@jrbourbeau 👋 same!

PR #212, opened 2022-03-15
Pin distributed

jacobtomlinson:
There seem to be some CI failures related to `distributed` importing something from `dask` that doesn't exist.

We should get things up to date here with latest versions of Dask, but as a short term fix I'm going to try and lock things down to the old versions that worked.


jsignell:
Superceded by #218

PR #213, opened 2022-04-15
Update Xarray atop example reference

jawjay:
Changed reference from dask.array.atop to dask.array.blockwise because the "da.atop function has moved to da.blockwise."


jrbourbeau (approved):
Thanks @jawjay!

jrbourbeau:
Also, I noticed this is your first code contribution to this repository. Welcome!

PR #187, opened 2021-03-23
[WIP] Update to JupyterLab 3

jrbourbeau:
cc @ian-r-rose

PR #215, opened 2022-04-18
Update `dask` and `distributed` version

jrbourbeau:
This updates to a more recent version of `dask` / `distributed` to avoid errors like

```python
 ---------------------------------------------------------------------------
ImportError                               Traceback (most recent call last)
Input In [1], in <cell line: 1>()
----> 1 from distributed import Client
      2 import dask_ml.datasets
      3 import dask_ml.ensemble

File /usr/share/miniconda3/envs/dask-examples/lib/python3.8/site-packages/distributed/__init__.py:3, in <module>
      1 from . import config
      2 from dask.config import config
----> 3 from .actor import Actor, ActorFuture
      4 from .core import connect, rpc
      5 from .deploy import LocalCluster, Adaptive, SpecCluster, SSHCluster

File /usr/share/miniconda3/envs/dask-examples/lib/python3.8/site-packages/distributed/actor.py:6, in <module>
      3 import threading
      4 from queue import Queue
----> 6 from .client import Future, default_client
      7 from .protocol import to_serialize
      8 from .utils import sync

File /usr/share/miniconda3/envs/dask-examples/lib/python3.8/site-packages/distributed/client.py:31, in <module>
     29 from dask.core import flatten, get_dependencies
     30 from dask.optimization import SubgraphCallable
---> 31 from dask.compatibility import apply
     32 from dask.utils import ensure_dict, format_bytes, funcname
     34 from tlz import first, groupby, merge, valmap, keymap, partition_all

ImportError: cannot import name 'apply' from 'dask.compatibility' (/usr/share/miniconda3/envs/dask-examples/lib/python3.8/site-packages/dask/compatibility.py)
ImportError: cannot import name 'apply' from 'dask.compatibility' (/usr/share/miniconda3/envs/dask-examples/lib/python3.8/site-packages/dask/compatibility.py)
```

which are [showing up in CI](https://github.com/dask/dask-examples/runs/6063390720?check_suite_focus=true)

PR #217, opened 2022-05-03
Update to JLab 3

ian-r-rose:
Supersedes #187.


binder/start
```diff
@@ -1,9 +1,13 @@
 #!/bin/bash
 
 # Replace DASK_DASHBOARD_URL with the proxy location
-sed -i -e "s|DASK_DASHBOARD_URL|${JUPYTERHUB_BASE_URL}user/${JUPYTERHUB_USER}/proxy/8787|g" binder/jupyterlab-workspace.json
+sed -i -e "s|DASK_DASHBOARD_URL|${JUPYTERHUB_BASE_URL}user/${JUPYTERHUB_USER}/proxy/8787|g" binder/example.jupyterlab-workspace
 
+# TODO: Importing of workspace via CLI is currently broken.
+# remove this workaround once https://github.com/jupyterlab/jupyterlab/issues/10805 is resolved
+mkdir -p ~/.jupyter/lab/workspaces/
+mv binder/example.jupyterlab-workspace ~/.jupyter/lab/workspaces/default-37a8.jupyterlab-workspace
```
jrbourbeau, line 9: The `37a8` part of this path seems quite specific. Why is that part needed? 

```suggestion
mv binder/example.jupyterlab-workspace ~/.jupyter/lab/workspaces/default-37a8.jupyterlab-workspace
```

jrbourbeau (commented):
Thanks @ian-r-rose! I've left a couple of questions but will defer to you on all things JupyterLab : )

The new `binder/example.jupyterlab-workspace` workspace file looks very similar (though differently formatted) to the old `binder/jupyterlab-workspace.json` file. What's the meaningful difference between the two?

jrbourbeau:
FYI there are unrelated CI failures on `main` (xref https://github.com/dask/dask-examples/pull/215), so I don't anticipate CI will pass on this PR

binder/start
ian-r-rose: This is 100% a hack -- that's a hash based on the workspace name which is not guaranteed to be stable. But I think it might be okay for the time-being, and is better than the currently broken state.

jrbourbeau: Gotcha -- totally fine with me if it works

ian-r-rose:
> What's the meaningful difference between the two?

There might not be a meaningful difference between them. Certainly one of the problems that I ran into is that some of the dashboard panels have changed names, but it depends on the specific layout.

ian-r-rose:
> some of the dashboard panels have changed names

Which reminds me: we should check this again once we update `distributed`

bryanwweber:
@jrbourbeau Yes, I'm working on completing #215

binder/environment.yml
```diff
@@ -7,9 +7,8 @@ dependencies:
   - dask=2.20.0
   - dask-image=0.2.0
   - dask-ml=1.6.0
-  - dask-labextension=2.0.2
-  - jupyterlab=2.1
-  - nodejs=16
+  - dask-labextension>=5.2.0
+  - jupyterlab>=3
```
bryanwweber, line 11: This introduces some conflicts with #218, so this should be merged first then I can fix that one.

jrbourbeau (approved):
Thanks @ian-r-rose -- let's give this a try!

PR #218, opened 2022-05-03
Update dependencies and examples

### LINKED ISSUE #206: Update dependencies and ensure all notebooks are working

bryanwweber:
This updates all the pinned dependencies to the current version as of today, and fixes any examples that were broken by the updates.

Supersedes #215. Closes #206.

**Important change:** This also adds all of the dependencies that I noticed that were required to run the examples and removes any `!pip install` cells in the Notebooks. This makes the binder image larger, so this may not be desirable. Happy to revert that change.

cc/ @jrbourbeau @ian-r-rose


dataframes/01-data-access.ipynb
```diff
@@ -11,27 +11,13 @@
   },
```
bryanwweber, line 11: <pre class="ql-syntax" spellcheck="false">Line #2.    client = Client(n_workers=1, threads_per_worker=4, processes=True, memory_limit='2GB')
</pre><p>This change is needed due to https://github.com/dask/dask/issues/8581</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/dask/dask-examples/pull/218/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='3'/>

applications/satellite-imagery-geotiff.ipynb
```diff
@@ -18,7 +8,7 @@
     "\n",
     "This notebook shows how to perform simple calculations with a GeoTIFF dataset using XArray and Dask. We load and rescale a Landsat 8 image and compute NDVI (Normalized difference vegetation index). This can be used to distinguish green vegetation from areas of bare land or water.\n",
     "\n",
-    "We'll use an image of the Denver, USA area taken in July 2018."
+    "We'll use an image of the Denver, USA area taken in July 2018. _Note: As of May 2022, these data are no longer publicly available from S3._"
```
jrbourbeau, line 11: Hmm that's unfortunate. Does this notebook still run then? 

I wonder if this is still publicly available, but just moved its location. There are Pangeo examples like https://gallery.pangeo.io/repos/pangeo-data/landsat-8-tutorial-gallery/landsat8.html that use Landsat 8 data. Perhaps we can update this example based on the Pangeo example notebook. Alternatively, if this notebook no longer runs we could temporarily not build/render it

jrbourbeau (commented):
Thanks for working on this @bryanwweber! 

> Important change: This also adds all of the dependencies that I noticed that were required to run the examples and removes any !pip install cells in the Notebooks. This makes the binder image larger, so this may not be desirable. Happy to revert that change.

How much longer does it take for the binder instances to spin up when including these dependencies? Generally, I like the idea of including everything that's needed in the conda environment file, but historically this has led to long binder spinup times that we'd like to avoid.

applications/satellite-imagery-geotiff.ipynb
ian-r-rose: I agree, if this doesn't run, I think we should remove/hide it.

dataframe.ipynb
```diff
@@ -258,7 +267,15 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "df[['x', 'y']].resample('24h').mean().compute().plot()"
+    "%matplotlib inline\n",
```
ian-r-rose, line 270: Are these still necessary? I haven't needed to run `%matplotlib inline` in a long time.

dataframes/01-data-access.ipynb
```diff
@@ -11,27 +11,13 @@
...
-   "source": [
-    "from IPython.display import HTML\n",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "from IPython.display import YouTubeVideo\n",
     "\n",
-    "HTML('<iframe width=\"560\" height=\"315\" src=\"https://www.youtube.com/embed/0eEsIA0O1iE?rel=0&amp;controls=0&amp;showinfo=0\" frameborder=\"0\" allowfullscreen></iframe>')"
+    "YouTubeVideo(\"0eEsIA0O1iE\")"
```
ian-r-rose, line 20: Nice

ian-r-rose (commented):
Thanks for this @bryanwweber!

applications/async-web-server.ipynb
```diff
@@ -139,7 +135,6 @@
    "cell_type": "code",
    "execution_count": null,
    "metadata": {
-    "collapsed": false,
```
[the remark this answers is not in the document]
ian-r-rose: I think it should default to `false`, so I suspect it will be okay

ian-r-rose:
>  Generally, I like the idea of including everything that's needed in the conda environment file, but historically this has led to long binder spinup times that we'd like to avoid.

In my experience, actually *pinning* the dependencies to specific versions results in faster build times. In call cases, I think the spin-up times for pre-built images won't differ that much. So I'd be in favor of getting an environment that works, then going to a running binder and doing something like `conda export --no-builds` to make it *more* specific.

applications/async-web-server.ipynb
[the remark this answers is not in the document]
bryanwweber: I ran all the notebooks through `jupyter nbconvert --clear-output --inplace`, which I suspect is the reason for these bulk changes.

dataframe.ipynb
bryanwweber: I don't know if they're needed anymore, I didn't intentionally add or remove any lines like this.

applications/satellite-imagery-geotiff.ipynb
bryanwweber: In the interest of finishing something during our time here, I opted to remove the example. I presume it could be updated and restored by someone who knows a bit about where this data can be accessed these days.

bryanwweber:
@jrbourbeau @ian-r-rose CI _finally_ all passed, so this is ready for another round of review

applications/embarrassingly-parallel.ipynb
```diff
@@ -384,7 +384,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "We encourage you to watch the [dashboard's status page](../proxy/8787/status) to watch on going computation."
+    "We encourage you to watch the [dashboard's status page](http://127.0.0.1:8787) to watch on going computation."
```
jrbourbeau, line 387: Does the previous link no longer work?

applications/forecasting-with-prophet.ipynb
```diff
@@ -17,20 +17,6 @@
...
-   "execution_count": null,
-   "metadata": {},
-   "outputs": [],
-   "source": [
-    "# This example currently relies on Prophet master\n",
-    "import subprocess\n",
-    "\n",
-    "subprocess.call([\n",
-    "    \"pip\", \"install\", \"prophet Cython cmdstanpy==0.9.5 pystan numpy pandas matplotlib LunarCalendar convertdate holidays setuptools-git python-dateutil tqdm\"\n",
-    "])"
```
jrbourbeau, line 31: Happy to see this go away : )

conf.py
```diff
@@ -50,7 +50,7 @@
 
 .. admonition:: Live Notebook
 
-    You can run this notebook in a `live session <https://mybinder.org/v2/gh/dask/dask-examples/main?urlpath=lab/tree/{{ docname }}>`_ |Binder| or view it `on Github <https://github.com/dask/dask-examples/blob/main/{{ docname }}>`_.
+    You can run this notebook in a `live session <https://mybinder.org/v2/gh/dask/dask-examples/main?urlpath=lab/tree/{{ docname }}>`__ |Binder| or view it `on Github <https://github.com/dask/dask-examples/blob/main/{{ docname }}>`__.
```
jrbourbeau, line 53: Just for my own understanding, why do we need a double underscore here?

index.rst
```diff
@@ -23,13 +23,16 @@ You can run these examples in a live session here: |Binder|
...
+   resilience
 
 .. toctree::
    :maxdepth: 1
    :caption: Dataframes
 
    dataframes/01-data-access
    dataframes/02-groupby
+   dataframes/03-from-pandas-to-dask
+   dataframes/04-reading-messy-data-into-dataframes
```
jrbourbeau, line 35: Nice! Thanks for adding these

jrbourbeau (commented):
Thanks for all your work on this @bryanwweber. Left a few comments, but overall this looks really close. 

Also, can we confirm that that the changes here don't increase binder startup times? It sounds like it will actually decrease them, which would be great

.github/workflows/ci-build.yaml
```diff
@@ -1,41 +1,55 @@
...
-          pip install nbsphinx dask-sphinx-theme>=2 sphinx
+          mamba install nbconvert nbformat jupyter_client ipykernel nbmake \
+          pytest nbsphinx dask-sphinx-theme>=2 sphinx
 
       - name: Execute Notebooks
-        shell: bash -l {0}
         run: |
-          pip install nbmake==0.1 pytest-xdist
           pytest \
-            -n=auto \
```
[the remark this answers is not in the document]
bryanwweber: Building notebooks in parallel puts warnings into the built docs that servers exist on 8787. I didn't think that was a good look, so I removed that for this build. The "update dependencies" build still goes in parallel for speed.

.github/workflows/ci-build.yaml
```diff
@@ -1,41 +1,55 @@
...
+        shell: bash -l {0}
     steps:
       - name: Checkout source
         uses: actions/checkout@v2
 
+      - name: Cache conda
+        uses: actions/cache@v2
+        env:
+          # Increase this value to reset cache if binder/environment.yml has not changed
+          CACHE_NUMBER: 0
```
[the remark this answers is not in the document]
bryanwweber: I think that's right.

.github/workflows/ci-build.yaml
```diff
@@ -1,41 +1,55 @@
...
       - name: Setup Conda Environment
         uses: conda-incubator/setup-miniconda@v2
         with:
-          miniconda-version: "latest"
+          miniforge-variant: Mambaforge
+          miniforge-version: latest
           environment-file: binder/environment.yml
           activate-environment: dask-examples
           auto-activate-base: false
+          use-only-tar-bz2: true
```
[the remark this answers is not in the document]
bryanwweber: This is required for the cache to work, https://github.com/conda-incubator/setup-miniconda#caching-packages

.github/workflows/ci-update-dependencies.yaml
```diff
@@ -0,0 +1,51 @@
...
+      - name: Execute Notebooks
+        run: |
+          pytest \
+            -vv \
+            -n=auto \
+            --forked \
+            --nbmake \
+            --overwrite \
+            --ignore machine-learning/torch-prediction.ipynb \
+            --ignore applications/json-data-on-the-web.ipynb
```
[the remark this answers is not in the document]
bryanwweber: There are now two environment files, `environment.yml` which has pinned versions created by `mamba env export --no-builds`, and `environment-base.yml` which only pins a few dependencies. This CI build uses `environment-base.yml` to create the environment, so it will (eventually) use updated versions of dependencies. The intention is to try to catch failures before users run into them.

.github/workflows/ci-update-dependencies.yaml
```diff
@@ -0,0 +1,51 @@
+name: Update Dependencies
```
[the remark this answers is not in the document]
bryanwweber: Well, yeah, "Update" as a verb is not appropriate... probably `Test with Updated Dependencies`

applications/embarrassingly-parallel.ipynb
bryanwweber: I'm not sure, Sphinx warns that the file doesn't exist.

conf.py
bryanwweber: The double underscore creates an anonymous link, rather than a link that can be reused. https://docutils.sourceforge.io/docs/user/rst/quickref.html#indirect-hyperlink-targets

bryanwweber:
> Also, can we confirm that that the changes here don't increase binder startup times? It sounds like it will actually decrease them, which would be great

I'm not sure how I can check this before this PR is merged. Is there a link I can use somewhere?

jrbourbeau:
https://mybinder.org/v2/gh/bryanwweber/dask-examples/fix-examples?urlpath=lab

jrbourbeau (approved):
Just tried launching a binder instance using this PR and it took ~1.5 minutes to pull the (already built) image and launch the JupyterLab session. Unfortunately I've not been able to compare with what we currently have on examples.dask.org as every time I've launched a binder session the image has already been present on the binder machine instance, so no need to pull it, and things launch pretty much instantly. That said, ~1.5 minutes to pull the image that corresponds to this PR seems within reason for what I would call "normal" binder launch times, so let's go ahead and merge this in.

Thanks again @bryanwweber!

PR #219, opened 2022-05-03
Update google analytics id

scharlottej13:
I'm fairly certain dask examples aren't being tracked in Google Analytics, and the google analytics id UA-18218874-5 doesn't appear to be attached to any properties currently being tracked on [analytics.dask.org](analytics.dask.org).

This is a draft because I'm not sure if other pieces need to be updated as well. The snippet from Google Analytics looks a bit different from what's here currently:

```html
<!-- Global site tag (gtag.js) - Google Analytics -->
<script async src="https://www.googletagmanager.com/gtag/js?id=UA-18218894-16"></script>
<script>
  window.dataLayer = window.dataLayer || [];
  function gtag(){dataLayer.push(arguments);}
  gtag('js', new Date());

  gtag('config', 'UA-18218894-16');
</script>
```

### DIFF
--- a/conf.py
+++ b/conf.py
@@ -67,7 +67,7 @@ nbsphinx_epilog = """
    m=s.getElementsByTagName(o)[0];a.async=1;a.src=g;m.parentNode.insertBefore(a,m)
    })(window,document,'script','https://www.google-analytics.com/analytics.js','ga');
 
-   ga('create', 'UA-18218874-5', 'auto');
+   ga('create', 'UA-18218894-16', 'auto');
    ga('send', 'pageview');
    </script>
    <!-- End Google Analytics -->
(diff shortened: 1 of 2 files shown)


mrocklin:
Hrm, I can't find the old asset anywhere .  @jacobtomlinson or @TomAugspurger was this you?  I'd like to get people access to this system's analytics.

Merging tomorrow to move this over to the main GA asset if no one speaks up.

TomAugspurger:
I was going to ask the same question as James :)

FWIW, under my analytics.google.com I see "Analytics Accounts" for dask-ml with ID UA-18218894-14, which is under an "mrocklin" Analytics account.

mrocklin:
Yeah, most of these are under my personal account (sorry!).  However the one for dask-examples isn't.  It's some other account.  Possibly it's some old thing of mine that I've lost track of.  If that's the case I'd like to centralize it on the new one.  However if it's in someone else's account then we might want to do something else (but maybe not?)  I wanted to check in with folks and see if anyone has context here. 

So far it looks like no one has any context.  I'll go ahead and merge this in a bit so that we can start getting very early numbers while we're together this week.

PR #222, opened 2022-06-06
Escape `dask-sphinx-theme` version with quotes

charlesbluca:
Noticed that the examples page hadn't been bumped to the new theme yet; think it's because the `dask-sphinx-theme` version isn't being quote-escaped in the `mamba install` command; this PR resolves this.

Will also scan through to see if there are any other places where this is happening.

### DIFF
--- a/.github/workflows/ci-build.yaml
+++ b/.github/workflows/ci-build.yaml
@@ -38,7 +38,7 @@ jobs:
       - name: Install testing and docs dependencies
         run: |
           mamba install nbconvert nbformat jupyter_client ipykernel nbmake \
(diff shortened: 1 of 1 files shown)


charlesbluca:
> we need quotes now as before https://github.com/dask/dask-examples/pull/221 we didn't need quotes

it looks like this has been an issue before #221 - looking at the logs of earlier runs the `mamba` command was affected in the same way. Still, I'm interested in why the interpreted install (i.e. `mamba install dask-sphinx-theme`) didn't go for 3.0.0 if that was the latest version.

> Would you mind adding a extra step

Sure, I can add that after the `mamba install step`

jrbourbeau (approved):

PR #225, opened 2022-07-27
Clarify difference between compute and load in xarray example

jsignell:
Some minor changes to the xarray tutorial that hopefully make it a little clearer. I'll merge this tomorrow if there are no comments.


jsignell:
Well it's not "tomorrow", but better late than never :)

Issue #220, opened 2022-05-04
Automatically clear notebook output


jrbourbeau:
Over in https://github.com/dask/dask-examples/pull/218 @bryanwweber mentioned he used `jupyter nbconvert --clear-output --inplace` to automatically clear the output from the example notebooks (xref https://github.com/dask/dask-examples/pull/218#discussion_r864965302). It would be great if there was some way we could automatically have this happen, for example in a `pre-commit` hook or something, to avoid accidentally committing notebooks that have output already

jsignell:
Looks like we can use [nbstripout](https://github.com/kynan/nbstripout/) like:

```python
repos:
- repo: https://github.com/kynan/nbstripout
  rev: 0.6.0
  hooks:
    - id: nbstripout
 ```

PR #226, opened 2022-07-27
Add pre-commit that strips out cells

jsignell:
This just adds the pre-commit as suggested by #220 - it would be good to also have a github action that enforces the pre-commit.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,12 @@
+repos:
+  - repo: https://github.com/asottile/pyupgrade
+    rev: v2.31.0
+    hooks:
+      - id: pyupgrade
+        args:
+          - --py38-plus
+  - repo: https://github.com/kynan/nbstripout
+    rev: 0.6.0
+    hooks:
+      - id: nbstripout
+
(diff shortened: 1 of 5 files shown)


jrbourbeau (commented):
+1 thanks @jsignell

> it would be good to also have a github action that enforces the pre-commit.

Agreed. Would you mind adding that workflow here? I imagine we can just reuse what's already over in `dask/dask`

.pre-commit-config.yaml
```diff
@@ -0,0 +1,12 @@
+repos:
+  - repo: https://github.com/asottile/pyupgrade
+    rev: v2.31.0
+    hooks:
+      - id: pyupgrade
+        args:
+          - --py38-plus
```
jsignell, line 7: I'm kind of of the opinion the more the merrier with pre-commit hooks and this one was easy to add.

jrbourbeau: +1. We're also already using this pre-commit in `dask/dask` and `dask/distributed` -- so I like the consistent code styling

jsignell: right. We can add black and things when we are feeling more ambitious :)

jrbourbeau (approved):

PR #238, opened 2023-11-23
Fix ``arbittrary`` --> ``arbitrary`` typo

habi:
arbittrary --> arbitrary


jrbourbeau (approved):
Thanks for catching this one @habi!

Issue #46, opened 2018-09-26
Dask Bag examples


mrocklin:
We currently lack dask bag examples in this repository.  Two come to mind:

1.  Read JSON data, and do some groupby aggregation with both `Bag.groupby` and `Bag.foldby`
2.  Read text data and do some basic wordcount

For the JSON data it might make sense to add a dataset generation tool for nested records data, similar to `dask.datasets.timeseries`, and then use that to generate JSON data to disk, similar to how we generate CSV data in http://examples.dask.org/dataframes/01-data-access.html#Create-artificial-dataset.

We would then read the JSON data, and do some minimal processing.

For the text data I wonder if there is an online dataset we can download.  I suspect that the complete works of shakespeare is around somewhere.  We might do a simple thing like read, split, frequencies.   Or we might do more complex work afterwards by bringing in NLTK, stemming words, removing stopwords, etc..

enstenr:
For point#1 i have raised a PR 
https://github.com/dask/dask-examples/pull/244

PR #244, opened 2025-11-24
 Added example for bag groupby and foldby as requested in issue 46

srajesh2712:
Added examples for using bag groupby and bag foldby  as per point#1 in below issue 
https://github.com/dask/dask-examples/issues/46

Added two example codes. 
Have provided comment inline as per the standard followed in other cells.

### DIFF
--- a/bag.ipynb
+++ b/bag.ipynb
@@ -139,6 +139,25 @@
    ]
   },
   {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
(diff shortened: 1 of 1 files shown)


srajesh2712:
@jacobtomlinson  could you take a look at this whenever possible.

jacobtomlinson:
I've merged a couple of text tweaks recently, but I'm not comfortable reviewing/merging code changes until we get CI unblocked #246

Issue #246, opened 2025-12-01
CI failing on main


jacobtomlinson:
CI is failing to run as the pinned software environments can no longer be reproduced. It's been a few years since these environments were updated. It also looks like there were some problems updating the solve in #232.

To get CI unblocked again we need to update the pinned dependencies following the [instructions in the README](https://github.com/dask/dask-examples#updating-the-binder-environment) and then make sure CI can build the notebooks, containers, docs, etc.

PR #248, opened 2026-03-12
Update ci-build.yaml

lvsltz:
Addresses #246 maybe?


.github/workflows/ci-build.yaml
jacobtomlinson: I think this needs bumping too as it calls cache internally

```suggestion
        uses: conda-incubator/setup-miniconda@v3
```

jacobtomlinson:
Thanks for opening this. I've nudged the PR along a little by updating all the actions versions and fixing up any migrations. The next blocker is that the pinned conda environment is no longer solvable. Could you take a look at this?

lvsltz:
Followed the instructions in the README and updated the environment.

